PR #2, opened 2024-05-31
Improved Evaluations

farzadab:
This PR does the following:
- Bugfix: AnyInstruct was not using val/train splits (as it had somehow gotten removed from the last PR)
- Faster evaluations by using DDP multiprocessing for text generation
- Validation loss added by default to training

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -353,16 +353,18 @@ class AnyInstructDataset(VoiceDataset):
         # The last 7 samples are missing audio files, so we exclude them.
         NUM_SAMPLES = 108193 - 7
         super().__init__(args)
-        dataset = (
-            datasets.load_dataset(
-                "json",
-                "anyinstruct",
-                data_files="https://huggingface.co/datasets/fnlp/AnyInstruct/resolve/main/speech_conv/metadata.jsonl",
-                split="train",
-            ).select(range(NUM_SAMPLES))
-            # TODO: make num_shards configurable if need be
-            .to_iterable_dataset(num_shards=16)
+        dataset = datasets.load_dataset(
+            "json",
+            "anyinstruct",
+            data_files="https://huggingface.co/datasets/fnlp/AnyInstruct/resolve/main/speech_conv/metadata.jsonl",
+            split="train",
+        ).select(range(NUM_SAMPLES))
+        dataset = dataset.train_test_split(
+            test_size=0.01, seed=args.shuffle_seed, shuffle=True
         )
+        dataset = dataset["train" if args.split == DatasetSplit.TRAIN else "test"]
+        # TODO: make num_shards configurable if need be
+        dataset = dataset.to_iterable_dataset(num_shards=16)
         if args.shuffle:
             dataset = dataset.shuffle(seed=args.shuffle_seed)
         self._init_dataset(dataset)
--- a/ultravox/training/config_base.py
+++ b/ultravox/training/config_base.py
@@ -25,6 +25,7 @@ class TrainConfig:
     data_dir: Optional[str] = None
     mds: bool = False
     num_samples: Optional[int] = None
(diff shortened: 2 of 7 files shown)


ultravox/training/train.py
```diff
@@ -170,21 +173,42 @@ def main() -> None:
...
+                    processor=processor,
+                    train_on_inputs=args.train_on_inputs,
+                ),
+                num_samples=num_samples,
+            )
+            for split_args, num_samples in [
+                (data_args, args.num_samples),
+                (val_data_args, args.val_num_samples),
+            ]
+        ]
```
farzadab, line 203: This looks ugly, I agree. Let me know if you have a better idea. Other solutions required more copy&pasting.

ultravox/training/ddp_utils_test.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+import torch.distributed
+from torch import multiprocessing as mp
+
+from ultravox.training import ddp_utils
+
+os.environ["MASTER_ADDR"] = "localhost"
+os.environ["MASTER_PORT"] = "12355"
```
farzadab, line 9: I needed to add some stuff to make it look like I'm running with `torchrun`.

ultravox/data/datasets.py
```diff
@@ -353,16 +353,18 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
...
+        dataset = datasets.load_dataset(
+            "json",
+            "anyinstruct",
+            data_files="https://huggingface.co/datasets/fnlp/AnyInstruct/resolve/main/speech_conv/metadata.jsonl",
+            split="train",
+        ).select(range(NUM_SAMPLES))
+        dataset = dataset.train_test_split(
+            test_size=0.01, seed=args.shuffle_seed, shuffle=True
         )
+        dataset = dataset["train" if args.split == DatasetSplit.TRAIN else "test"]
```
juberti, line 365: should this just be `dataset[args.split]`?

ultravox/training/config_base.py
```diff
@@ -54,7 +55,7 @@ class TrainConfig:
     optimizer: str = "adamw_torch"
     num_epochs: int = 1
     max_steps: int = 0
-    eval_steps: Optional[int] = None
+    val_steps: Optional[int] = None
```
juberti, line 58: just checking, did you mean to replace `eval_steps` (as `eval_num_samples` wasn't replaced)

ultravox/training/train.py
juberti: maybe just hoist this logic into a `prepare_dataset` function that takes `args` and `num_samples`

juberti (approved):

ultravox/training/config_base.py
farzadab: No. The idea is that we have 3 things now: training, validation, final (generative) evaluation.

HF calls the second one `eval` (e.g. `eval_dataset`, `eval_steps`), but I wanted to make sure the names `val` refers to validation and `eval` refers to the final generative evaluation.

On that note, I see that I need to rename `eval_accum_steps` to `val_accum_steps` as well to keep things consistent.

ultravox/data/datasets.py
farzadab: The issue is we have "validation" and "train" but the `train_test_split` function creates "test" and "train" splits.

farzadab:
Small training run to check for regressions seems to have run fine: https://wandb.ai/fixie/ultravox/runs/q3rh3yxv

ultravox/training/train.py
farzadab: Refactored. The parameter list is a bit long, but I prefer that over either sending the whole train_args or using a nested function.

juberti: looks great

PR #6, opened 2024-06-04
Optionally include "passage" in BoolQ samples

juberti:
This turns BoolQ from an internal knowledge task to a reading comprehension task, and I believe is the right way to run this eval.


farzadab (approved):
LGTM.

PR #12, opened 2024-06-06
Add tts_tool, for converting a HF dataset to audio

juberti:
e.g., 
```
just tts -d google/boolq -c question -a audio -u fixie-ai/boolq-audio
```


ultravox/tools/tts.py
```diff
@@ -0,0 +1,48 @@
...
+    return f"""
+    <speak version="1.0" xml:lang="en-US">
+        <voice xml:lang="en-US" name="{voice}">
+            {saxutils.escape(text)}
+        </voice>
+    </speak>"""
+
+
+class AzureTts:
+    DEFAULT_VOICE = "en-US-JennyNeural"
```
farzadab, line 21: Is there a way to randomize this or cycle through multiple voices?

ultravox/tools/tts_tool.py
```diff
@@ -0,0 +1,73 @@
+import argparse
+import os
+
+import datasets
+
+from ultravox.tools import tts
+
+# This script is used to generate audio samples from text using a TTS model.
+# Ex: python tts_tool.py -d google/boolq -c question -a audio -u "fixie-ai/boolq-audio
+parser = argparse.ArgumentParser()
```
farzadab, line 10: nit: you could use simple_parsing instead for better typing

ultravox/tools/tts_tool.py
```diff
@@ -0,0 +1,73 @@
+import argparse
+import os
+
+import datasets
+
+from ultravox.tools import tts
+
+# This script is used to generate audio samples from text using a TTS model.
+# Ex: python tts_tool.py -d google/boolq -c question -a audio -u "fixie-ai/boolq-audio
```
farzadab, line 9: Does that mean we can incrementally add other fields too if need be? For example convert passage as well?

ultravox/tools/tts_tool.py
```diff
@@ -0,0 +1,73 @@
...
+    col_name: str,
+    audio_col_name: str,
+):
+    def _tts_batch(batch):
+        batch[audio_col_name] = [
+            {"bytes": tts_client.tts(text)} for text in batch[col_name]
+        ]
+        return batch
+
+    return ds_split.map(_tts_batch, batched=True).cast_column(
```
farzadab, line 35: nit: [default `batch_size`](https://huggingface.co/docs/datasets/v2.19.0/en/package_reference/main_classes#datasets.Dataset.map) seems to be 1000, right? Or did I read that wrong?

I guess things are fine and there aren't too many requests if it worked?

farzadab (approved):

ultravox/tools/tts_tool.py
juberti: Yep, this is just a v1. We could perhaps allow specifying multiple columns to be converted in a follow up PR.

farzadab: Yeah I meant this already enables that which is neat!

ultravox/tools/tts_tool.py
juberti: The requests are currently all serialized since we're not using asyncio or any multiprocessing for the TTS requests, so the batch size doesn't matter all that much. But it would be useful to rework this in a follow up PR, we currently can only process ~10 rows per second with the current approach.

ultravox/tools/tts.py
juberti: Not at present. We could allow -V to accept either a string or a list though in a follow up PR.

PR #14, opened 2024-06-07
Tool for adding new synthetic columns

farzadab:
This tools allows you to add new columns, by using a template and feeding it into `gpt-4o` to create a synthetic "ground-truth" response.

image

```
$ just enrich_ds -d fixie-ai/boolq-audio -s train -b with_explanation -c explanation -t $HF_WRITE_TOKEN

Loading dataset "fixie-ai/boolq-audio", new column name: "explanation", template:
Passage: {passage}

Question: {question}

Answer: {answer}

Provide a short explanation to the question given the passage that entails the answer.
Processing split "train"...

Map (num_proc=16): 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9427/9427 [16:47<00:00,  9.35 examples/s]

### CODE REVIEW

File: ultravox/tools/enrich_synth_data.py [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+        token = args.token or os.environ.get("HF_TOKEN")
+
+        try:
+            new_split.push_to_hub(
+                args.upload_name,
+                config_name=args.dataset_subset,
+                split=split,
+                token=token,
+                revision=args.upload_branch,
+            )
```
farzadab, line 89: This kept failing for me. I think the right approach is to apply this to all splits and upload the final `data_dict` but I didn't try it.

juberti: Do you have a HF write token?

juberti (commented):
This is neat! I kind of think we should integrate this into tts_tool (or maybe rename that to ds_tool) since most of the code seems to be the same. Wdyt?

ultravox/tools/enrich_synth_data.py
farzadab: Yes I do. The issue was that the schema was changing and it didn't allow changing the schema one-split at a time.

farzadab:
I agree that it's best to integrate these two into the same file. I'll take a look to see if there's a way to combine them in a simplified manner.

ultravox/tools/enrich_synth_data.py
juberti: ah ok, makes sense.

farzadab:
PTAL

ultravox/tools/enrich_synth_data.py
farzadab: I didn't verify if this works, but I'm now uploading the both splits at the same time. I think that would fix the issue since the splits won't be out of sync at any point in time.

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+import os
+from typing import Any, Dict, Optional, Union
+
+import datasets
+import openai
+import simple_parsing
+
+from ultravox.tools import tts
+
+CHAT_CLIENT = openai.Client()
```
juberti, line 11: I don't think ALL_CAPS is the right naming for object like this. I think `chat_client` or `tts_client` on their own are the right choice here. Alternatively, maybe you could push these objects into their respective tasks?

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+TTS_CLIENT = tts.AzureTts()
+
+DEFAULT_TEXTGEN_TEMPLATE = """Passage: {passage}
+
+Question: {question}
+
+Answer: {answer}
+
+Provide a short explanation to the question given the passage that entails the answer."""
```
juberti, line 20: "entails" is an unusual choice here. Maybe "Given the question and passage, provide a short explanation for the provided answer" or some such. 

Also, this is fine for now, but it seems hard to have a useful default for this sort of thing.

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+    upload_name: Optional[str] = simple_parsing.field(default=None, alias="-u")
+    upload_branch: Optional[str] = simple_parsing.field(default="main", alias="-b")
+    num_shards: Optional[int] = simple_parsing.field(default=None, alias="-N")
+
+    token: Optional[str] = simple_parsing.field(default=None, alias="-t")
+
+    task: Union[TtsTask, TextGenerationTask] = simple_parsing.subgroups(
+        {"tts": TtsTask, "textgen": TextGenerationTask},  # type: ignore
+        default_factory=TtsTask,
+        alias="-T",
```
juberti, line 61: Ideally this wouldn't need an argument name, we could do this as a positional parameter (e.g., `ds_tool tts`) Doesn't need to be solved in this PR though if it's tricky.

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+        },
+    ).cast_column(audio_col_name, datasets.Audio(sampling_rate=task.sample_rate))
+
+
+def _text_gen_sample(sample, template: str, col_name: str, language_model: str):
+    input_text = template.format(**sample)
+    response = CHAT_CLIENT.chat.completions.create(
+        model=language_model,
+        messages=[{"role": "user", "content": input_text}],
+        max_tokens=128,
```
juberti, line 98: probably going to want to customize these in the future, I'd make them properties of the task

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+    sample, col_name: str, audio_col_name: str, voice: str, sample_rate: int
+):
+    text = sample[col_name]
+    text = text["text"] if isinstance(text, dict) else text
+    sample[audio_col_name] = TTS_CLIENT.tts(text, voice=voice, sample_rate=sample_rate)
+    return sample
+
+
+def _tts_split(
+    ds_split: datasets.Dataset, task: TtsTask, num_proc: Optional[int] = None
```
juberti, line 75: num_proc can never be `None` so it doesn't need to be `Optional`

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+        messages=[{"role": "user", "content": input_text}],
+        max_tokens=128,
+        temperature=0,
+    )
+    sample[col_name] = response.choices[0].message.content
+    return sample
+
+
+def _text_gen_split(
+    ds_split: datasets.Dataset, task: TextGenerationTask, num_proc: Optional[int] = None
```
juberti, line 106: num_proc can never be `None` so it doesn't need to be `Optional`

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+    ds_split: datasets.Dataset, task: TtsTask, num_proc: Optional[int] = None
+):
+    col_name = task.column_name
+    audio_col_name = task.audio_column_name or f"{col_name}_audio"
+    print(f'TTS mapping "{col_name}" to "{audio_col_name}"...')
+
+    return ds_split.map(
+        _tts_sample,
+        num_proc=num_proc,
+        fn_kwargs={
```
juberti, line 84: FWIW, I liked the local function approach that I had before as it avoided having to curry all these parameters to the map function

ultravox/tools/ds_tool.py
juberti: If we keep the param name though I'd make the alias "-t" for `task` and "-T" for `template` and remove the alias for `token`.

ultravox/tools/ds_tool.py
```diff
@@ -0,0 +1,172 @@
...
+    audio_column_name: Optional[str] = simple_parsing.field(default=None, alias="-a")
+    voice: Optional[str] = simple_parsing.field(default=None, alias="-V")
+    sample_rate: int = simple_parsing.field(default=16000, alias="-r")
+
+
+@dataclasses.dataclass(frozen=True)
+class TextGenerationTask:
+    language_model: str = simple_parsing.field(default="gpt-4o", alias="-m")
+    new_column_name: str = simple_parsing.field(default="explanation", alias="-c")
+    template: str = DEFAULT_TEXTGEN_TEMPLATE
```
juberti, line 35: don't need to worry about this for now but I think we should allow @foo.txt syntax to load the template from a file (similar to curl @data.json)

juberti (approved):
Nice work on this, turned out great.

ultravox/tools/ds_tool.py
farzadab: > "entails" is an unusual choice here.

You might be right.
Just to explain though, the idea is to be able to use it so that Ultravox outputs: `{explanation} \n Answer: {answer}`.  In this way, we want the generated "explanation" be able to "precede" the answer. That's why I wrote the prompt like so. Text generated via the proposed prompt might cause the wording to be slightly incorrect.

> it seems hard to have a useful default for this sort of thing.

I agree. I didn't know where to put the info though. For Expresso, I put it in the comments. I could alternatively put this info in the dataset README as well, but that might not be enough.

ultravox/tools/ds_tool.py
farzadab: I'll see if I can do that. The `frozen=True` might cause issues for adding the object in those dataclasses.

ALL_CAPS: I thought that was the Google style, but maybe I'm wrong?
So global variables (not constant values like number/string), don't need to be ALL_CAPS?
image

ultravox/tools/ds_tool.py
farzadab: I feel like that approach adds complexity without any value:
1. we needed to handle the threading ourselves, which is bug prone.
2. we needed to handle the batched version of the code which for us provided no value, just needless complexity (batching is faster when you do things like vector math, but we'd just be batching and unbatching needlessly).

ultravox/tools/ds_tool.py
farzadab: btw I'm not attached to the prompt though. If you think the new version is better given the explanation above then I'll change it.

ultravox/tools/ds_tool.py
juberti: oh yeah super glad to see the threading done internally, just wondering if we can use a local func to avoid having to pass all these `fn_kwargs` since the local func will have access to the local vars directly

farzadab: I assumed nested functions were usually a no/no:
image

Alternatively, I was going to give it just the task.
image

But if you don't see any issues with nested functions, then I can switch over.

juberti: well, you're doing that exact thing - closing over a local value other than self/cls - so I think it's valid. That said I don't feel super strongly about this

ultravox/tools/ds_tool.py
juberti: I'm still a bit confused regarding how the explanation is to be used. But mainly I was reacting to "entails" and so maybe a smaller change would be "entails" -> "provides a rationale for"

ultravox/tools/ds_tool.py
farzadab: You're right, I read the "other than" incorrectly there.
Okay using nesting instead.

juberti (approved):

ultravox/tools/ds_tool.py
juberti: See if it actually makes it simpler - the use of `task` definitely helps a lot already, so use your own judgement as to whether the local functions are actually still useful.

ultravox/tools/ds_tool.py
farzadab: Done.

ultravox/tools/ds_tool.py
farzadab: Both look similar. Not a big deal either way it can be changed later on if need be.

PR #16, opened 2024-06-07
Local tokenizer and processor for more consistent CI

farzadab:


ultravox/inference/infer_test.py
```diff
@@ -14,13 +14,15 @@
 @pytest.fixture(scope="module")
 def tokenizer():
     return transformers.AutoTokenizer.from_pretrained(
-        "meta-llama/Meta-Llama-3-8B-Instruct"
+        "../tests/assets/hf/Meta-Llama-3-8B-Instruct"
```
juberti, line 17: probably don't need to have this path at the root. Suggest using
`assets/hf/foo` (i.e., put the assets folder under the top-level `ultravox`)

ultravox/inference/infer_test.py
```diff
@@ -14,13 +14,15 @@
...
     )
 
 
 @pytest.fixture(scope="module")
```
juberti, line 21: add a comment explaining why we use local files, for posterity

juberti (approved):
CI on this PR seems to work, so that's good!

juberti:
Still seeing failures here even with local tokenizer/processor. I think something is getting wedged during test shutdown, kind of frustrating to debug since everything works fine locally 🙁

juberti:
OK, so the issue here was ddp_utils_test.py, not infer_test.py. That said, I think this is worth keeping since it allows external PRs to access the Llama 3 tokenizer, which would otherwise fail due to not having the HF_TOKEN set.

PR #22, opened 2024-06-11
model: Add initial `HubertWithKmeans` module and extract tool.

shaper:
For use in extracting discrete HuBERT tokens from audio files.

Adapted from https://github.com/lucidrains/audiolm-pytorch

### DIFF
--- a/ultravox/model/hubert.py
+++ b/ultravox/model/hubert.py
@@ -1,0 +1,119 @@
+from pathlib import Path
+
+import fairseq
+import joblib
+import torch
+from einops import rearrange
+from einops import repeat
+from torch import nn
+from torchaudio.functional import resample
+
+_DEFAULT_OUTPUT_LAYER = 9
+_DEFAULT_TARGET_SAMPLE_HZ = 16000
+# From the paper, 20 ms at 16 kHz.
+_DOWNSAMPLE_FACTOR = 320
+
+
+def _round_down_nearest_multiple(num, divisor):
+    return num // divisor * divisor
+
+
+def _curtail_to_multiple(t, mult, from_left=False):
+    data_len = t.shape[-1]
+    rounded_seq_len = _round_down_nearest_multiple(data_len, mult)
+    seq_slice = (
+        slice(None, rounded_seq_len) if not from_left else slice(-rounded_seq_len, None)
+    )
+    return t[..., seq_slice]
+
+
+def _exists(val):
+    return val is not None
+
+
+class HubertWithKmeans(nn.Module):
+    """
+    Adapted from @lucidrains
+    [implementation](https://github.com/lucidrains/audiolm-pytorch/tree/main).
+
+    Original note:
+
+    Checkpoint and kmeans can be downloaded at
+    https://github.com/facebookresearch/fairseq/tree/main/examples/hubert or you
+    can train your own
+    """
+
+    def __init__(
+        self,
+        checkpoint_path,
+        kmeans_path,
+        target_sample_hz=_DEFAULT_TARGET_SAMPLE_HZ,
+        seq_len_multiple_of=None,
+        output_layer=_DEFAULT_OUTPUT_LAYER,
+    ):
+        super().__init__()
+
+        self.target_sample_hz = target_sample_hz
+        self.seq_len_multiple_of = seq_len_multiple_of
+        self.output_layer = output_layer
+
+        model_path = Path(checkpoint_path)
+        kmeans_path = Path(kmeans_path)
+
+        assert model_path.exists(), f"path {checkpoint_path} does not exist"
+        assert kmeans_path.exists(), f"path {kmeans_path} does not exist"
+
+        checkpoint = torch.load(checkpoint_path)
+        load_model_input = {checkpoint_path: checkpoint}
+        model, *_ = fairseq.checkpoint_utils.load_model_ensemble_and_task(
+            load_model_input
+        )
+
+        self.model = model[0]
+        self.model.eval()
+
+        kmeans = joblib.load(kmeans_path)
+
+        self.kmeans = kmeans
+
+        self.register_buffer(
+            "cluster_centers", torch.from_numpy(kmeans.cluster_centers_)
+        )
+
+    @property
+    def groups(self):
+        return 1
+
+    @property
+    def codebook_size(self):
+        return self.kmeans.n_clusters
+
+    @property
+    def downsample_factor(self):
+        return _DOWNSAMPLE_FACTOR
+
+    @torch.inference_mode()
+    def forward(self, wav_input, flatten=True, input_sample_hz=None):
+        if _exists(input_sample_hz):
+            wav_input = resample(wav_input, input_sample_hz, self.target_sample_hz)
+
+        if _exists(self.seq_len_multiple_of):
+            wav_input = _curtail_to_multiple(wav_input, self.seq_len_multiple_of)
+
+        embed = self.model(
+            wav_input,
+            features_only=True,
+            mask=False,  # thanks to @maitycyrus for noticing that mask is defaulted to True in the fairseq code
+            output_layer=self.output_layer,
+        )["x"]
+
+        batched_cluster_centers = repeat(
+            self.cluster_centers, "c d -> b c d", b=embed.shape[0]
+        )
+        dists = -torch.cdist(embed, batched_cluster_centers, p=2)
+        clusters = dists.argmax(dim=-1)
+
+        if flatten:
+            return clusters
+
+        return rearrange(clusters, "b ... -> b (...)")
--- a/ultravox/tools/extract_hubert.py
+++ b/ultravox/tools/extract_hubert.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
+from pathlib import Path
+
+import numpy as np
+import requests
+import torch
+from loguru import logger
+from torchaudio import load
+from torchaudio.functional import resample
+from tqdm import tqdm
+
+from ultravox.model.hubert import HubertWithKmeans
+
+_AUDIO_SUFFIXES = ("*.wav", "*.mp3", "*.flac", "*.ogg")
+_MULTILINGUAL_OUTPUT_LAYER = 11
+_SAMPLE_RATE = 16000
+
+_SCRIPT_ROOT_DIR = Path(__file__).parent.parent
+_MODEL_DIR = _SCRIPT_ROOT_DIR / "models"
+
+# English
+_MODEL_CHECKPOINT = "hubert_base_ls960.pt"
+_MODEL_KMEANS = "hubert_base_ls960_L9_km500.bin"
+
+# Multilingual
+_MODEL_CHECKPOINT_MULTI = "mhubert_base_vp_en_es_fr_it3.pt"
+_MODEL_KMEANS_MULTI = "mhubert_base_vp_en_es_fr_it3_L11_km1000.bin"
+
+
+def _download_file(filename):
+    save_path = _MODEL_DIR / filename
+    if save_path.exists():
+        logger.debug(f"Skipping download of existing file [file={save_path}]")
+        return
+
+    logger.debug(f"Downloading [file={save_path}]")
+    response = requests.get(
+        f"https://dl.fbaipublicfiles.com/hubert/{filename}", stream=True
+    )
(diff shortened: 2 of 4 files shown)


shaper:
Hi there, welcome to the first PR and thank you for your consideration and steering as we explore together! A few notes as I begin to find my way:
- initially I had a PR with only the `hubert.py` model. However, it felt bare without the tool to show usage. Open to breaking apart if you'd prefer.
- have you seen https://github.com/Delgan/loguru and are you open to trying it out? If you'd prefer I can swap in `logging` instead (we've been using `loguru` for a while and much prefer it, by default for example it will log the source file/method and line number for log statements). It's only in the tool, so you can try running it to see what it looks like (`loguru` not yet included in `requirements.txt`, you'll need to install it).
- I put the new deps in `requirements-dev.txt` since , in our explorations we can allow for cpu-based token extraction from dataset samples at runtime (e.g. not precomputed) which isn't really `dev/tools` at that point. Not sure we'd want to allow that, though, as it'd slow the datasets and adds complexity. So, just mentioning as placement reqs-file-wise isn't 100% clear.
- testing: I started a unit test for `HubertWithKmeans` but, lol -- how best to test an `nn.Module`, one might ask. All the unit test really did was mock out the `torch`, `fairseq`, etc. and allow the ctor to pass, which isn't particularly useful beyond validating the ctor signature. More useful would be something like a trivial integration test that runs it with the tool on a few files, or at least the model on a fixture audio file, etc. There isn't infra for this today and given it's experimental and so forth it felt like overkill, so I backed out the test work and figured we'd discuss. Similarly, the tool is pretty simple and I'm not sure it warrants tests yet. I'm open to adding whatever you'd like, and can help set up fixture/etc. bits as needed, now or later.
- the tool has to download the HuBERT checkpoint and kmeans data and store it locally. I didn't see clear precedence for this, so I added a new `models` dir to repo root and to `.gitignore`, but open to thought. I know typically one uses HF to download pretrained models and the HF cache handles it, however I don't believe the exact files we want here are set up for this on HF today (which doesn't mean we couldn't put them there, so again, open to feedback).

juberti:
The HF thing was the first thing that jumped out at me - I tend to think we'll be happier if we have a unified way of hosting and downloading models. If this turns into a bear we can of course revisit.

Minor nits:
- please add mypy types
- please follow Google Python style guide and import only packages, not individual classes

ultravox/model/hubert.py
```diff
@@ -0,0 +1,119 @@
+from pathlib import Path
+
+import fairseq
```
farzadab, line 3: In general I'm skeptical of new requirements that are not standalone and `fairseq` is a quite heavy requirement with its own PyTorch requirements.

Can't we use the HuggingFace version instead? I don't see KMeans there, but that should be easy to add separately.

ultravox/tools/extract_hubert.py
```diff
@@ -0,0 +1,119 @@
...
+    logger.info(f"Found {len(audio_files)} audio files.")
+    for filename in tqdm(
+        audio_files, desc="Extracting Hubert tokens", total=len(audio_files)
+    ):
+        _process_file(device, hubert, filename)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.add_argument("data_dir")
```
farzadab, line 116: I'm not fully opposed to offline creation of the dataset, but we'll definitely have to balance the pros and cons. I see 3 options generally for this:

1. online streaming (which we currently use in Ultravox)
2. offline done once to update dataset
3. offline every time at the start of training

I'm not sure if you had 2 or 3 in mind.
- Option 3 doesn't make much sense to me (at least as slow as 1 and causes training runs to be started with a huge delay).
- Option 2 is mainly good when you are fairly certain which encoder you want to use, which IMO is still up to debate at this point.
- Option 1 gives you the most flexibility. If you have any data that shows the costs for this is too high I can be convinced to give up that flexibility

farzadab (commented):
Thanks for the PR Walter!

I'm still quite unsure about what are the changes that you had in mind. Maybe it would be good to have a meeting.

The things that I'd like to know:
1. what datasets were used and with what prompts
2. how long did training take and it would be nice to see some training curves
3. what are the material changes in training required other than the addition of HuBert tokens for output

farzadab:
Actually I know Justin said to separate out the PR, but I'd really like to see the entire changes first.
Just the diff (maybe with your `main` and ours) could be good enough.

ultravox/model/hubert.py
```diff
@@ -0,0 +1,119 @@
...
+    return t[..., seq_slice]
+
+
+def _exists(val):
+    return val is not None
+
+
+class HubertWithKmeans(nn.Module):
+    """
+    Adapted from @lucidrains
```
farzadab, line 36: Q: Does this mean Copyright notice should be copied over?
Mainly asking because the copyright stuff is very foreign to me, so if you know we don't need to I'd love to learn!

juberti: IANAL, but that repo is also MIT licensed, so I don't think any additional notices need to be copied over.

shaper: Yes, it's MIT so it should be fine to use. I think the way licenses/credits are handled varies by project/repo. Sometimes rigor around putting third party derived code in a separate dir with a LICENSE file, etc. I think given the early stage of this project, one lightweight idea could be to start a `CREDITS.md` file or similar and as new items come up, add a bullet on directories/files, link to source, and state the license.

ultravox/tools/extract_hubert.py
shaper: This work is still early and experimental for sure. I had in mind to land the experimental tools and such as we go, presuming they get to some point where it seems likely they'll be of at least semi lasting value. However, if we are thinking of this repo as more production-ready oriented work or want to separate things that are still super early into other subdirectories, etc., I am happy to shift my thinking.

Here I had 2 in mind, though in our experimentation we've also done 1. It did appear to slow training, I want to say fairly substantially, but I'll look at it again and follow up.

In general where we are today is we're looking to effectively update various datasets to add HuBERT tokens (and I imagine SNAC and others as well over time) and it seems advantageous from an overall compute perspective to have updated datasets with those tokens. Even if they live only locally with folks experimenting, vs. pushed to HF or otherwise, having a tool to allow doing this seems useful. We are using it today e.g. I used it for AnyInstruct and am looking now at what other datasets could be most useful for experimenting with, and will apply it to those as well.

ultravox/model/hubert.py
shaper: Let me take another look at the HF version. It didn't look like it had the multilingual variant, and it was a different checkpoint, and given this evolved script we've been working with for a bit is operational (and I wasn't sure yet how you all are thinking about managing models/binaries, via HF or not), I figured I'd just send this for discussion. I agree it'd be simpler if we can just use that.

shaper:
> The things that I'd like to know:
> 
> 1. what datasets were used and with what prompts
> 2. how long did training take and it would be nice to see some training curves
> 3. what are the material changes in training required other than the addition of HuBert tokens for output

Let's follow up soon and happy to have a discussion. We are not yet as far as you describe above though we aim to move very quickly -- we are actively working on proper training next. What we've done so far is roughly:
- add audio_hubert_tokens to voice samples
- increase the model token space to allow for hubert tokens
- repurpose the `audio` tag to populate it with said hubert tokens
- trivial training with a straightforward test prompt that produces said tokens, so as to:
- ensure that the ultravox model can ingest hubert tokens and produce them at inference time
- ensure that our acoustic/vocoder models can take those tokens and produce quality speech

There are various directions we can go in editing/adding training/datasets with the goal to teach the model to be able to speak HuBERT for responses while preserving existing functionality. I expect it could take time to find the one we like best, and so that we'll be playing around with things here quite a bit. We're optimistic that we can quickly identify some initial training run/dataset changes that are not large in scope and support speech output, and we'd value your input and steering on approach!

Similar to my other comment, with the above as the planned direction, having a model/tool in repo to allow generating the tokens (and iteratively cleaning up and landing some of the bulleted work above, e.g. behind a flag or otherwise generalized/abstracted) seems useful to support a team effort for what's likely long-lived exploration.

Maybe another way to put it all is -- we have several of the likely foundational pieces together or in mind (semantic speech tokens, voice model/vocoder), we're learning the Ultravox internals/tooling and starting to fit them together, and we've done work to de-risk/validate and make sure the end to end flow is feasible and roughly operational. Another key upcoming work item re: foundational pieces is to also include our models for inference purposes.

I will put together a PR to sketch the above bullet points (done-so-far) soon but I don't expect all of it is yet something we'd want to land. This tool/model's case was clearer (but I understand your point re: using existing HF model if it fits).

ultravox/tools/extract_hubert.py
farzadab: Okay makes sense. Assuming that you're treating this as an experimental feature and the original path still remains viable then I'm more than fine with this.

farzadab:
Thanks for the explanation.
I think I understand the disconnect here. I thought you had training working already (which was totally surprising to me).

Repeating to make sure I understand, what you did was some sort of "overfitting" experiment just to make sure it was possible to get the tokens out. Not really a full-fidelity trained model that could respond in voice as it saw fit.

I'm aligned with the direction now.

### Production vs experimentation and how to handle them:
Until now I've generally tended to do highly experimental features in a branch and only push them to main after getting more confident, but that becomes less viable as more people want to collaborate on the same path.

In some projects we had an `experimental` or `projects` directory under `main` for each person/project idea. This directory had looser constraints (e.g. no/less testing, possibly different Python requirements). However I don't think that's necessary for us yet.

I guess my rule of thumb is: highly experimental features related to training/model need to be proven at least somewhat first before being merged, but in your case you've already done the due diligence so we're good.

And as a rule, I'm open to experimental features as long as the existing paths still remain viable and not highly impacted.

juberti:
Agree with @farzadab. This is definitely part of the roadmap so I don't think we need to worry too much about exactly what directory this lives in, but I do want to be cautious about taking on large dependencies and also align on core concepts like how we load models.

I definitely know that the PR process isn't always the easiest way to communicate, so don't hesitate to post/DM if we're giving unclear/conflicting advice.

ultravox/model/hubert.py
shaper: After looking around the HuBERT parts we need are quite scattered on HF and IMO our best path is to host a copy of the data we want in either Ultravox or LMNT HF repo, let me know how you'd like to handle it (if Ultravox do I have access, should I point you to binaries to upload yourself, etc.).

Summary of findings:

- the canonical references for English and Multilingual (with kmeans bin as well) are not on HF, they are at:
  - https://github.com/facebookresearch/fairseq/tree/main/examples/hubert
  - https://github.com/facebookresearch/fairseq/blob/main/examples/speech_to_speech/docs/textless_s2st_real_data.md
- there are as you saw some facebook HF HuBERT repos but they only have the one base model file, only English, no km-bin (a non-Facebook HF repo does have the multilingual km-bin), and the model appears to be a different format
- As an elaboration for background as useful whether now or future, from discussion with @sharvil. we explicitly don't want to use the fine-tuned HuBERT model variants. Our intent here is to use HuBERT tokens as a proxy for phonemes and preserve more pronunciation/filler words/etc. The fine-tuned HuBERT models are a different token space and AIUI effectively overly-lossy for our needs.
- I hear you on `fairseq` being heavyweight, it looks like it's really just used to load the model so I'll see if I can port to straight torch and will follow up shortly. There is some logic in their utility routine but it doesn't seem too bananas.

farzadab: Thanks for checking this out.

If you feel like the lift is gonna be too much work, then we can just go with `fairseq` as well. My intention wasn't to block you.

Uploading to HF is also a good idea. Given the messy state of things, I think many people will appreciate the gesture.
If you do end up creating the straight PyTorch version, I think pushing to HF hub can be done as a follow up.

I don't have a strong preference as to where it should live, specially since you're the one doing the work. @juberti can probably give you access to our org as well.

shaper:
Still working on this, also getting familiar with poetry, lol! Removing fairseq would be yet further helpful as it depends on `hydra-core` which breaks testing for some reason. More tomorrow.

shaper:
> Repeating to make sure I understand, what you did was some sort of "overfitting" experiment just to make sure it was possible to get the tokens out. Not really a full-fidelity trained model that could respond in voice as it saw fit.

Yes, that's right. Working on the trained model as well in parallel.

PR #24, opened 2024-06-11
Switch from Pip to Poetry

### LINKED ISSUE #18: Migrate pip setup to Poetry
The whole `activate` process is quite clunky, let's just sidestep all of this and let Poetry manage the venv.

juberti:
Fixes #18


mcloud.yaml
```diff
@@ -1,17 +1,16 @@
...
+  cluster: r7z22p1
 integrations:
   - integration_type: git_repo
     git_repo: fixie-ai/ultravox
     git_branch: $UV_BRANCH
-    pip_install: -r requirements.txt
+    pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/$UV_CONFIG
+  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/$UV_CONFIG
```
farzadab, line 13: Nice!
Did you try a training run with this?
Does it work and also does it make installation stage slower?

juberti: I was able to start a run, I didn't wait for it to finish though. AFAICT it doesn't make install any slower (it's essentially pip but with an auto-managed VM).

The only thing I haven't verified is whether it munges the existing `torch` install. If that's an issue, it will be easy enough to put `torch` in its own group and simply use `poetry install --without torch`

farzadab: Yeah I mainly meant `torch`.
And yeah I can see Poetry being used in [the run](https://wandb.ai/fixie/ultravox/runs/obhfzykg/overview).

farzadab (approved):

PR #26, opened 2024-06-12
datasets: Handle converting `int16` audio data in `VoiceSample`.

shaper:
We saw `VoiceSample` failing the assert on `float32` audio data when playing around with the Gradio infer app and submitting an `mp3` file. We didn't dig deeper into Gradio (I'm sure it's possible to alter/convert there as well), but it seems potentially useful for `VoiceSample` to handle `int16` audio data on top of what it already handles.

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -159,6 +159,10 @@ class VoiceSample:
         if self.audio is not None:
             if self.audio.dtype == np.float64:
                 self.audio = self.audio.astype(np.float32)
+            elif self.audio.dtype == np.int16:
+                self.audio = self.audio.astype(np.float32) / np.float32(32768.0)
+            elif self.audio.dtype == np.int32:
+                self.audio = self.audio.astype(np.float32) / np.float32(2147483648.0)
             assert (
                 self.audio.dtype == np.float32
             ), f"Unexpected audio dtype: {self.audio.dtype}"
@@ -166,7 +170,7 @@ class VoiceSample:
 
     messages: List[Dict[str, str]]
     """List of messages, each with a "role" and "content" field."""
-    audio: Optional[np.ndarray] = None
+    audio: Optional[np.typing.NDArray[np.float32]] = None
     """Audio data as float32 PCM @ `sample_rate`."""
     sample_rate: int = SAMPLE_RATE
     """Audio sample rate in Hz."""
--- a/ultravox/data/datasets_test.py
+++ b/ultravox/data/datasets_test.py
@@ -1,8 +1,9 @@
 import itertools
-from typing import Optional
+from typing import Optional, Union
 
 import datasets as hf_datasets
 import numpy as np
+import pytest
 import torch
 from torch.utils import data
 from transformers.feature_extraction_utils import BatchFeature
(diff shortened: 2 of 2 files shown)


ultravox/data/datasets.py
```diff
@@ -159,6 +159,11 @@ def __post_init__(self):
         if self.audio is not None:
             if self.audio.dtype == np.float64:
                 self.audio = self.audio.astype(np.float32)
+            elif self.audio.dtype == np.int16:
+                # Gradio mp3 file submission, as one example, seems to turn an
+                # mp3 into int16, thus it can be useful to convert to float32 if
+                # we come across this.
```
farzadab, line 165: Nit: I vote to remove the comment. It makes sense for the PR, but I don't think it makes as much sense once the PR has been merged.

ultravox/data/datasets.py
```diff
@@ -159,6 +159,11 @@ def __post_init__(self):
...
+                self.audio = self.audio.astype(np.float32) / np.float32(32768.0)
```
farzadab, line 166: I'm wondering if this fix obviates the need for these lines:
https://github.com/fixie-ai/ultravox/blob/main/ultravox/inference/infer.py#L70_L71

farzadab (approved):
LGTM, thanks!

ultravox/data/datasets.py
farzadab, line 166: Seems like the issue is not limited to `int16`. We can get `int32` as well:
```suggestion
                self.audio = self.audio.astype(np.float32) / np.float32(32768.0)
            elif self.audio.dtype == np.int32:
                self.audio = self.audio.astype(np.float32) / np.float32(2147483648)

```

farzadab: Of course another solution is to change Gradio input side:
`gr.Audio(label="Audio", show_download_button=True)` -> add `type="filepath` and load the audio file ourselves.

ultravox/data/datasets.py
juberti: Thanks for explaining the rationale though, I was curious where you were hitting this.

ultravox/data/datasets.py
juberti: Yeah, and possibly also in UltravoxProcessor too.

juberti (approved):
If it's easy, suggest adding a test to datasets_test.py, a la the existing `create_sample`.

ultravox/data/datasets.py
shaper: Would you prefer I look at removing the two you mention as part of this PR, or separately? Probably separately is better, but asking while I learn your PR preferences.

juberti: nah, go ahead and land this PR, definitely possible that cleaning up those other lines starts pulling on a thread...

shaper: > Of course another solution is to change Gradio input side: `gr.Audio(label="Audio", show_download_button=True)` -> add `type="filepath` and load the audio file ourselves.

There seem to be enough code paths that can lead into this that it's worth adding your suggestion. It may also be worth the Gradio change but if so leaving for another time.

shaper: Will look at the Infer and Processor bits where we may be able to remove things next.

shaper:
> If it's easy, suggest adding a test to datasets_test.py, a la the existing `create_sample`.

Added tests, using some type hints in the test resulted in the need to enhance `VoiceSample.audio` type from `audio: Optional[np.ndarray] = None` to `audio: Optional[NDArray[np.float32]] = None` which I think is beneficial (and didn't create any other type issues elsewhere).

Will wait a bit to merge in case there's feedback on the tests.

juberti (approved):

shaper:
@juberti need you or @farzadab to land as I don't have write access. Excitement! Thanks for the onboarding guidance.

farzadab:
Merged and gave you write access. Thanks!

PR #28, opened 2024-06-13
Add @file syntax to ds_tool

juberti:
Also:
- makes the task name positional
- defaults subset to None, as not every default is "default"
- refactors the map func to be inside the task, making tasks more encapsulated

### DIFF
--- a/ultravox/tools/ds_tool.py
+++ b/ultravox/tools/ds_tool.py
@@ -32,17 +32,16 @@ class TtsTask:
             self.audio_column_name = f"{self.column_name}_audio"
 
     def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Dataset:
-        def map_sample(sample):
-            text = sample[self.column_name]
-            text = text["text"] if isinstance(text, dict) else text
-            sample[self.audio_column_name] = tts_client.tts(text)
-            return sample
-
         print(f'TTS mapping "{self.column_name}" to "{self.audio_column_name}"...')
-
-        return ds_split.map(map_sample, num_proc=num_proc).cast_column(
+        return ds_split.map(self._map_sample, num_proc=num_proc).cast_column(
             self.audio_column_name, datasets.Audio(sampling_rate=self.sample_rate)
         )
+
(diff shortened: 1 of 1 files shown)


farzadab (approved):
LGTM.

Please make sure to update `Justfile` as well.

ultravox/tools/ds_tool.py
```diff
@@ -41,16 +54,36 @@ class TextGenerationTask:
     max_tokens: int = 128
     temperature: float = 0
 
+    def __post_init__(self):
+        if self.template.startswith("@"):
+            self.template = open(self.template[1:], "r").read()
```
farzadab, line 59: nit: suggest using `with`. The `open(...).read()` pattern won't close the file.
```suggestion
            with open(self.template[1:], "r") as template_file:
                self.template = template_file.read()
```

ultravox/tools/ds_tool.py
```diff
@@ -31,6 +31,19 @@ def __post_init__(self):
         if self.audio_column_name is None:
             self.audio_column_name = f"{self.column_name}_audio"
 
+    def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Dataset:
+        def map_sample(sample):
+            text = sample[self.column_name]
+            text = text["text"] if isinstance(text, dict) else text
+            sample[self.audio_column_name] = tts_client.tts(text)
+            return sample
```
farzadab, line 39: nit: given that `self` is already present, there's no need to use a `nested` function here.

ultravox/tools/ds_tool.py
```diff
@@ -41,16 +54,36 @@ class TextGenerationTask:
     max_tokens: int = 128
     temperature: float = 0
 
+    def __post_init__(self):
+        if self.template.startswith("@"):
+            self.template = open(self.template[1:], "r").read()
+
+    def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Dataset:
+        def map_sample(sample):
```
farzadab, line 62: Same here.

PR #30, opened 2024-06-14
BoolQ for Training and Eval

farzadab:
This PR adds:
- Training version of BoolQ (extended with GPT-based "explanations")
- True/False (exact match) evaluations for BoolQ
- Text-only evaluations for BoolQ and AnyInstruct to show the text-audio gap

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -434,7 +434,7 @@ class AnyInstructInputDataset(AnyInstructDataset):
 
     def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
         audio_transcript = row["chat"][0]["message"]
-        return VoiceSample(
+        return self._make_sample(
             self._get_transcribe_messages(idx, audio_transcript),
             self._load_anyinstruct_audio(row["chat"][0]["speech"]),
             audio_transcript=audio_transcript,
@@ -447,7 +447,7 @@ class AnyInstructOutputDataset(AnyInstructDataset):
 
     def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
         audio_transcript = row["chat"][1]["message"]
-        return VoiceSample(
+        return self._make_sample(
             self._get_transcribe_messages(idx, audio_transcript),
             self._load_anyinstruct_audio(row["chat"][1]["speech"]),
             audio_transcript=audio_transcript,
@@ -456,11 +456,10 @@ class AnyInstructOutputDataset(AnyInstructDataset):
 
 class BoolQDataset(VoiceDataset):
     def __init__(self, args: VoiceDatasetArgs) -> None:
-        assert (
-            args.split == DatasetSplit.VALIDATION
-        ), f"BoolQ is only for validation, but got split={args.split}"
         super().__init__(args)
-        dataset = self._load_audio_dataset("fixie-ai/boolq-audio", split="train")
+        dataset = self._load_audio_dataset(
+            "fixie-ai/boolq-audio", split=args.split.value
+        )
         self._init_dataset(dataset)
 
     def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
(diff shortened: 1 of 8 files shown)


ultravox/data/datasets.py
```diff
@@ -192,7 +192,7 @@ class VoiceDatasetArgs:
     """If `prompt` is not set, the number of canned prompts to use."""
     include_audio: bool = True
     """Whether to include audio in the samples."""
-    include_context: bool = False
+    include_context: bool = True
```
farzadab, line 195: I didn't see a point in keeping this as False by default (iow I don't see many places where we want it to be False).

ultravox/data/datasets.py
```diff
@@ -479,6 +478,96 @@ def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
...
+    ANSWER_PROMPTS = [
+        "Answer: ",
+        "A: ",
+        "",
+        "The answer is: ",
+        "Result: ",
+        "Conclusion: ",
+    ]
+
+    def _get_query_prompt(self, idx: int) -> str:
```
farzadab, line 514: I went a bit crazy with the prompting here since without it I was getting huge overfitting.
I did not experiment with milder version of the prompting combination though.

ultravox/data/datasets.py
```diff
@@ -479,6 +478,96 @@ def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
...
+        Example prompt:
+            Passage: {context}
+
+            Question: {question}
+
+            Provide a short explanation, then respond with True/False on the last line.
+        """
+        if self._args.prompt:
+            return self._args.prompt
+        prompt_idx = idx % min(self._args.num_prompts, len(self.BOOLQ_PASSAGE_PROMPTS))
```
farzadab, line 526: @juberti you started the `idx % LEN` format, but for training we actually prefer not to tie each sample with a single prompt.

I think the original thinking was that we don't want to see any sample twice which is eventually what we want, so I'm not gonna change it.

ultravox/evaluation/eval_types.py
```diff
@@ -26,4 +26,12 @@ class WerResult:
     score: float
 
 
-Result = Union[InstructResult, WerResult]
+@dataclasses.dataclass
+class ExactMatchResult:
```
farzadab, line 30: Adding a class seemed a bit weird, but I went with it.

ultravox/training/evaluation.py
```diff
@@ -44,14 +45,36 @@ def dataset_infer(
...
+class EvalScenario:
+    name: str
+    dataset: str
+    metric: str
+    include_audio: bool = True
+    include_context: bool = True
+    new_tokens: Optional[int] = None
+
+
+EVAL_SCENARIOS = [
```
farzadab, line 58: Using the new dataclass allows us to move these scenarios into configs later on.
I was hesitant to do it yet though.

ultravox/training/configs/stage2_lora.yaml
```diff
@@ -1,16 +1,23 @@
...
+num_prompts: 11
 
 lr: 1.e-4  # need a lower LR for LLM fine-tuning
+lr_scheduler: constant_with_warmup
 lr_warmup_steps: 250
 max_steps: 5_000
+save_steps: 0
+
+batch_size: 2
+grad_accum_steps: 2
```
farzadab, line 23: When calculating the budget for SODA I noticed that some BoolQ passages are very long. It's possible that we can filter those out and keep `batch_size=4`.

TODO.

farzadab: IIUC, with Llama3 and Whisper, our token budget with `batch_size=4` is somewhere between 668 and 728.

We could definitely filter BoolQ to get rid of the few samples that fall in this regime (less than 1%).

With SODA, the ratio is higher though (depending on where we draw the line).

farzadab: Okay nevermind. The training did not get faster.
GPU utilization was already at > 90% and returning batch_size to 4 even somehow made things slower. Reverting to `batch_size=2 grad_acc_steps=2`.

farzadab: Marking PR ready to review.

ultravox/training/configs/stage2_lora.yaml
```diff
@@ -1,16 +1,23 @@
...
 # disable_layer_drop: True
 # audio_model_lora_config:
 #   r: 64
 #   target_modules: ['k_proj', 'q_proj', 'v_proj', 'out_proj', 'intermediate_dense', 'output_dense']
 
-num_prompts: 6
+num_prompts: 11
 
 lr: 1.e-4  # need a lower LR for LLM fine-tuning
+lr_scheduler: constant_with_warmup
```
juberti, line 17: why not cosine?

ultravox/training/evaluation.py
```diff
@@ -44,14 +45,36 @@ def dataset_infer(
...
+    EvalScenario("boolq__wer", "boolq_in", "asr"),
+    EvalScenario("anyinstruct__instruct_follow", "anyinstruct", "instruct"),
+    EvalScenario(
+        "boolq_binary", "boolq_passage", "exact_match_last_word", new_tokens=128
```
juberti, line 62: What's the thinking with one vs two underscores? Should this be `boolq__binary`?

ultravox/training/evaluation.py
juberti: Makes sense to me. Maybe alphabetize these scenarios?

ultravox/training/evaluation.py
```diff
@@ -44,14 +45,36 @@ def dataset_infer(
...
+    ),
+    EvalScenario(
+        "anyinstruct__instruct_follow_text_only",
+        "anyinstruct",
+        "instruct",
+        include_audio=False,
+    ),
+    EvalScenario(
+        "boolq_binary_text_only",
+        "boolq_passage",
```
juberti, line 72: does this dataset exist?

juberti (approved):

ultravox/data/datasets.py
farzadab: What was the "original concern"?

No I'm saying with multiple epochs, doing this is actually gonna be bad: we could be adding more diversity, but samples index `i` is always gonna be represented in the same way.

Not a big issue at all. Just though to raise it (maybe we find an easy solution or at least we're both aware of the pros/cons).

ultravox/data/datasets.py
```diff
@@ -434,7 +434,7 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
 
     def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
         audio_transcript = row["chat"][0]["message"]
-        return VoiceSample(
+        return self._make_sample(
```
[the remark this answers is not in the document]
farzadab: Hmm, let me take a look.
The reason I did this was that without it you couldn't be sure in which scenario you will get `text_only` and in which scenario you wouldn't. Silently refusing features in some cases seemed weird to me.

ultravox/data/datasets.py
```diff
@@ -479,6 +478,96 @@ def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
...
+        prompt = self.BOOLQ_PASSAGE_PROMPTS[prompt_idx]
+
+        # Separate either with 1 or 2 newlines, depending on idx
+        # 13, 17, 19 are prime numbers (to avoid a pattern)
+        separator = self.SEPARATORS[
+            idx % 13 % min(self._args.num_prompts, len(self.SEPARATORS))
+        ]
+
+        query_prompt = self.QUERY_PROMPTS[
+            idx % 17 % min(self._args.num_prompts, len(self.QUERY_PROMPTS))
```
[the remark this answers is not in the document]
farzadab: Okay so that goes back to my comment above: if you don't see an issue with doing that then I'm game.

I guess I should've asked my question is: why did you originally use `idx` pattern instead of `random`. I thought maybe it was due to issues with multi-processing/forking.

ultravox/training/configs/stage2_lora.yaml
farzadab: This is a very old change tbh, I just never pushed it.

Two reasons:
  1) with constant the val_loss metrics actually went lower
  2) I've always wondered if our models are under-trained, but with cosine, looking at the loss I kept thinking that training had almost converged, when in fact it had not.

Even if evidence for (1) wasn't there (and if in the future it reverses), I think (2) alone can cause issues as we are still exploring the subject. Maybe our final (stable) training runs should use cosine (if it turns out to be better), but not our exploratory runs.

Graphs for cosine (purple) vs constant (yellow):
image

You can see eval/loss going way lower with yellow, and train/loss is actually going up for some reason.

I then let the training run for much longer and the training did not converge even with 8x the amount of training.
I should note that the test metrics didn't support that training longer leads to much benefits so I haven't touched `max_steps` for our exploratory train jobs. The v0.2 model was trained for longer though.

ultravox/training/evaluation.py
farzadab: You're right, I missed that one.

> What's the thinking?

I was mainly trying to match the strings I had before.

ultravox/training/evaluation.py
farzadab: Good catch, I changed that last minute. Seems like I didn't change it everywhere.

ultravox/data/datasets.py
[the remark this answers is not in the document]
farzadab: Yeah okay other than the silent refusal, without this I wouldn't have been able to do the `anyinstruct__instruct_follow_text_only` evaluation.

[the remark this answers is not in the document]
juberti: shouldn't that eval have used `AnyInstructAnswerDataset`?

ultravox/data/datasets.py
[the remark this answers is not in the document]
juberti: I was worried that using `random` would make things non-reproducible, but with a private RNG and known seed I think things should be deterministic.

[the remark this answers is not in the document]
farzadab: I feel like the RNG change could be a bit more involved. If this is more of a nit, I'd rather land this and address that in a separate PR.

ultravox/data/datasets.py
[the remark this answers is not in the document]
farzadab: Hmm, you're right, it does. I don't think we use `AnyInstructInputDataset` and `AnyInstructOutputDataset` anywhere right now. They can be removed if need be.

What are your thoughts on the non-uniformity part?

ultravox/training/evaluation.py
farzadab: Alphabetized, but still kept the `text_only` evaluations separate.

ultravox/data/datasets.py
```diff
@@ -479,6 +478,96 @@ def _get_sample(self, idx: int, row: transformers.BatchFeature) -> VoiceSample:
...
+        """
+        if self._args.prompt:
+            return self._args.prompt
+        prompt_idx = idx % min(self._args.num_prompts, len(self.BOOLQ_PASSAGE_PROMPTS))
+        prompt = self.BOOLQ_PASSAGE_PROMPTS[prompt_idx]
+
+        # Separate either with 1 or 2 newlines, depending on idx
+        # 13, 17, 19 are prime numbers (to avoid a pattern)
+        separator = self.SEPARATORS[
+            idx % 13 % min(self._args.num_prompts, len(self.SEPARATORS))
```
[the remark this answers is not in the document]
farzadab: Will do this in the RNG PR.

ultravox/training/evaluation.py
farzadab: Also, `_` in my mind separates compound words, whereas `__` separates different concepts, so here it was: `{dataset}__{metric}[__text_only]`.

farzadab:
Okay I'm gonna put the randomization changes into a new PR. Other than that all comments should already have been addressed.

ultravox/data/datasets.py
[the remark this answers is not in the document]
farzadab: Okay I'm gonna go with the `_make_sample` version.

PR #31, opened 2024-06-14
Add Eleven and Fireworks support to ds_tool

juberti:
TTS impl is controlled by the `-i` argument.
LLM impl is controlled by the `-b` and `-k` arguments.
Also support random voices via the `-V random` argument.

### DIFF
--- a/ultravox/tools/ds_tool.py
+++ b/ultravox/tools/ds_tool.py
@@ -8,8 +8,8 @@ import simple_parsing
 
 from ultravox.tools import tts
 
-chat_client = openai.Client()
-tts_client = tts.AzureTts()
+tts_client: tts.Client
+chat_client: openai.Client
 
 DEFAULT_TEXTGEN_TEMPLATE = """Passage: {passage}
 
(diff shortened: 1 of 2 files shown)


ultravox/tools/ds_tool.py
```diff
@@ -50,10 +53,16 @@ class TextGenerationTask:
...
+    base_url: str = simple_parsing.field(
+        default="https://api.openai.com/v1", alias="-b"
+    )
+    api_key: str = simple_parsing.field(default=None, alias="-k")
     max_tokens: int = 128
     temperature: float = 0
 
     def __post_init__(self):
+        global chat_client
+        chat_client = openai.Client(base_url=self.base_url, api_key=self.api_key)
```
farzadab, line 65: nit: assigning to global vars seems like a bad smell but I know why you had to do it.
I suggest putting a comment here that explains the reasoning: e.g. multi-processing not allowing non-serializable objects.

ultravox/tools/tts.py
```diff
@@ -38,10 +88,64 @@ def tts(self, text: str, voice: Optional[str] = None, sample_rate: int = 16000):
...
+            },
+        }
+        return self._handle_pcm_response(self._post(url, headers, body))
+
+
+def create_client(implementation: str, sample_rate: int):
+    if implementation == "azure":
+        return AzureTts(sample_rate=sample_rate)
+    elif implementation == "eleven":
+        return ElevenTts(sample_rate=sample_rate)
```
farzadab, line 150: Nit: I'm wondering if we ever want to randomize between the two implementations. Not important though.

ultravox/tools/ds_tool.py
```diff
@@ -22,14 +22,17 @@
...
     column_name: str = simple_parsing.field(default="question", alias="-c")
     audio_column_name: Optional[str] = simple_parsing.field(default=None, alias="-a")
     voice: Optional[str] = simple_parsing.field(default=None, alias="-V")
     sample_rate: int = simple_parsing.field(default=16000, alias="-r")
 
     def __post_init__(self):
+        global tts_client
         if self.audio_column_name is None:
             self.audio_column_name = f"{self.column_name}_audio"
+        tts_client = tts.create_client(self.implementation, self.sample_rate)
```
farzadab, line 35: Same (nit) here: please add a comment about the need for global var usage.

farzadab (approved):
LGTM, thanks.

PR #32, opened 2024-06-15
Extending `ds_tool` for SODA conversational dataset

farzadab:


ultravox/tools/ds_tool.py
```diff
@@ -112,6 +136,8 @@ def main(args: DatasetToolArgs):
 
     for split, ds_split in data_dict.items():
         print(f'Processing split "{split}"...')
+        if args.shuffle:
```
juberti, line 139: can't we shuffle the dataset once it's created (ie in training?)

ultravox/tools/ds_tool.py
```diff
@@ -78,12 +96,18 @@ def _map_sample(self, sample):
 # Ex: just ds_tool tts -d google/boolq -u fixie-ai/boolq-audio -c question -a audio --token $HF_WRITE_TOKEN
 # Ex: just ds_tool textgen -d fixie-ai/boolq-audio -u fixie-ai/boolq-audio -c explanation
 # Ex: just ds_tool textgen -d ylacombe/expresso -u fixie-ai/expresso -c continuation -T @expresso_template.txt
+# Ex: just ds_tool textgen -d allenai/soda --shuffle True --split train -n 10000 -u fixie-ai/soda -c alt_last_turn -j -m llama3-8b -T "dialogue[:-1]"
+# Ex: just ds_tool textgen -d allenai/soda --shuffle True --split validation -n 1000 -u fixie-ai/soda -c alt_last_turn -j -m llama3-8b -T "dialogue[:-1]"
+# Ex: just ds_tool textgen -d allenai/soda --shuffle True --split test -n 1000 -u fixie-ai/soda -c alt_last_turn -j -m llama3-8b -T "dialogue[:-1]"
```
juberti, line 101: I think Jinja can do both the expression handling and formatting, e.g., `{{dialogue[:-1]}}`. Not having to have two separate modes would be nice...

ultravox/tools/ds_tool.py
farzadab: Agree that shuffle without `num_samples` doesn't make much sense, but for SODA we don't want to convert the whole set.

ultravox/tools/ds_tool.py
farzadab: With Jinja the result is going to be a string, not a list, right?
I don't see how I can cleanly then re-create the list of turns to give to the ChatCompletion API.

juberti: It's a string, of course, but it can be in JSON format and probably auto-converted to a sequence accordingly.
e.g.
```
template = Template('''
[
{% for person in data %}
  {
    "name": "{{ person.name }}",
    "age": {{ person.age }}
  }{% if not loop.last %},{% endif %}
{% endfor %}
]
''')
```

ultravox/tools/ds_tool.py
```diff
@@ -42,16 +37,18 @@ def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Datas
...
         sample[self.audio_column_name] = tts_client.tts(text, self.voice)
         return sample
 
 
 @dataclasses.dataclass
 class TextGenerationTask:
-    new_column_name: str = simple_parsing.field(default="explanation", alias="-c")
-    template: str = simple_parsing.field(default=DEFAULT_TEXTGEN_TEMPLATE, alias="-T")
+    new_column_name: str = simple_parsing.field(alias="-c")
+    template: str = simple_parsing.field(alias="-T")
```
farzadab, line 51: Moved the existing template out and I didn't see a point in keeping the default.

ultravox/tools/ds_tool.py
```diff
@@ -72,10 +69,20 @@ def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Datas
...
+        rendered = jinja2.Template(self.template).render(**sample)
+
+        try:
+            turns = json.loads(rendered)
+            assert isinstance(turns, list)
+            assert all(isinstance(turn, dict) for turn in turns)
+            assert len(turns) > 0
+            assert turns[-1].get("role", None) == "user"
+        except json.JSONDecodeError:
+            turns = [{"role": "user", "content": rendered}]
```
farzadab, line 81: I find the `try/except` pattern is quite weird, but that was the only way I found to allow for both simple (template for the message content only) and full on control of the `messages` list.

ultravox/tools/ds_tool.py
farzadab: Using Jinja now, but I had to use a slightly weird try/catch method for it. PTAL.

ultravox/tools/ds_tool.py
```diff
@@ -138,9 +148,13 @@ def main(args: DatasetToolArgs):
 
     try:
         if args.dataset_split:
-            data_dict[args.dataset_split].push_to_hub(args.upload_name, **hub_args)
-        else:
-            data_dict.push_to_hub(args.upload_name, **hub_args)
+            # load the full dataset, otherwise the existing splits will be overwritten
+            upload_ds = datasets.load_dataset(args.upload_name, args.dataset_subset)
```
juberti, line 152: should we just push the individual split in this case (not as a `DatasetDict`?)

ultravox/tools/tts.py
```diff
@@ -132,7 +134,7 @@ def tts(self, text: str, voice: Optional[str] = None):
             i = np.random.randint(len(self.ALL_VOICES)) + os.getpid()
             voice = self.ALL_VOICES[i % len(self.ALL_VOICES)]
         url = f"https://api.elevenlabs.io/v1/text-to-speech/{voice}/stream?output_format=pcm_16000"
-        print("url", url)
+        logging.debug(f"url {url}")
```
juberti, line 137: oops, I didn't mean to leave this in, you can remove this.

ultravox/tools/tts.py
```diff
@@ -145,9 +147,38 @@ def tts(self, text: str, voice: Optional[str] = None):
         return self._handle_pcm_response(self._post(url, headers, body))
 
 
+class CachedClientWrapper:
+    def __init__(self, client: Client, provider: str):
+        super().__init__()
+        self._client = client
+        self._base_path = os.path.join(".cache/ds_tool", provider)
```
juberti, line 154: maybe `.cache/ds_tool/tts`

ultravox/tools/tts.py
```diff
@@ -145,9 +147,38 @@ def tts(self, text: str, voice: Optional[str] = None):
...
+        text_hash = hashlib.sha256(text.encode()).hexdigest()
+        os.makedirs(path, exist_ok=True)
+
+        cache_path = os.path.join(path, f"{text_hash}.wav")
+
+        if os.path.exists(cache_path):
+            with open(cache_path, "rb") as f:
+                return f.read()
+
+        pcm = self._client.tts(text, voice)
```
juberti, line 167: Output is actually WAV rather than PCM, FWIW

ultravox/tools/tts.py
```diff
@@ -145,9 +147,38 @@ def tts(self, text: str, voice: Optional[str] = None):
         return self._handle_pcm_response(self._post(url, headers, body))
 
 
+class CachedClientWrapper:
```
juberti, line 150:
```suggestion
class CachingClientWrapper:
```

juberti (approved):
LG, just a few nits.

ultravox/tools/ds_tool.py
farzadab: Not sure if that accomplishes the same thing:

1. that would require having two arguments to represent the same thing: column + template and column is only used to look this thing up in this case :?
2. You can maybe have one column, but want to add a system prompt or other stuff anyway which would not be possible in this paradigm.


Alternative is to make this an explicit choice: for example, `-j` as in JSON mode vs text mode.

ultravox/tools/ds_templates/README.md
```diff
@@ -0,0 +1,33 @@
...
+```bash
+just ds_tool textgen -d allenai/soda --shuffle True -s test -n 1000 -u fixie-ai/soda -c alt_last_turn \
+    -T @ultravox/tools/ds_templates/soda_template.jinja -b https://api.fireworks.ai/inference/v1 \
+    -k $FIREWORKS_API_KEY -m accounts/fireworks/models/llama-v3-8b-instruct --token $HF_WRITE_TOKEN
+
+just ds_tool textgen -d allenai/soda --shuffle True -s validation -n 1000 -u fixie-ai/soda -c alt_last_turn \
+    -T @ultravox/tools/ds_templates/soda_template.jinja -b https://api.fireworks.ai/inference/v1 \
+    -k $FIREWORKS_API_KEY -m accounts/fireworks/models/llama-v3-8b-instruct --token $HF_WRITE_TOKEN
+
+just ds_tool textgen -d allenai/soda --shuffle True -s train -n 100000 -u fixie-ai/soda -c alt_last_turn \
```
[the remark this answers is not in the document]
farzadab: The point wasn't for this to be an example, it was to document the dataset creation process.
I agree that `scripts/` is a good alternative as well. Less ideal for documentation, but more ideal for actionable commands.

ultravox/tools/ds_tool.py
farzadab: I'll take another look. I think that was what I was doing before and it ended up not working correctly, but I could be wrong.

ultravox/tools/tts.py
farzadab: It shows that I still don't know what PCM is, no? :))
Renaming to wav.

ultravox/tools/ds_templates/README.md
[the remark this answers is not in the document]
farzadab: PTAL. I created two scripts under `scripts/dataset_creation`. As for README, I can remove it.

ultravox/tools/ds_tool.py
farzadab: You're right. The previous version somehow had missed the `split=...` argument, so the fix was way simpler.

ultravox/tools/ds_tool.py
farzadab: Added a `json_mode (-j)` instead to make the check explicit (i.e. no try/except needed).

ultravox/tools/wrappers.py
```diff
@@ -0,0 +1,59 @@
...
+        response = self._client.chat.completions.create(**kwargs)
+        text = response.choices[0].message.content
+
+        with open(cache_path, "w") as f:
+            f.write(text)
+
+        return text
+
+
+class CachingTtsWrapper:
```
farzadab, line 37: I had to add a cache for text generation as well, so I ended up moving both caching wrappers into the same file and out of `tts.py`.

ultravox/tools/ds_templates/template_test.py
```diff
@@ -0,0 +1,23 @@
+import jinja2
+import json
+
+
+def test_quotes():
```
farzadab, line 5: Added a replace function to SODA template as well as a test for it.

ultravox/tools/wrappers.py
```diff
@@ -0,0 +1,60 @@
...
+import openai
+
+from ultravox.tools import tts
+
+
+class CachingChatWrapper:
+    def __init__(self, client: openai.Client, base_url: str):
+        super().__init__()
+        self._client = client
+        self._base_path = os.path.join(
```
juberti, line 15: this probably should also include the model name too

ultravox/tools/ds_tool.py
juberti: I see what you are getting at, it's the output of the template that matters here and that could be either json or str depending on the template. Using -j to make this deterministic seems like a good strategy.

ultravox/tools/wrappers.py
```diff
@@ -0,0 +1,60 @@
+import hashlib
```
juberti, line 1: might want to call this `caching.py`. Also, don't need to do this now, but we're probably going to want to push ds_tool, tts, caching, templates down into their own ds_tool folder.

juberti (approved):

ultravox/tools/wrappers.py
farzadab: Yeah makes sense. Will do.

scripts/dataset_creation/soda_audio.sh
```diff
@@ -0,0 +1,24 @@
...
+    -k $FIREWORKS_API_KEY -m accounts/fireworks/models/llama-v3-8b-instruct --token $HF_WRITE_TOKEN
+
+just ds_tool textgen -d allenai/soda --shuffle True -s train -n 100000 -u fixie-ai/soda-audio -c alt_last_turn \
+    -T @ultravox/tools/ds_templates/soda_alt_last_turn.jinja -j -b https://api.fireworks.ai/inference/v1 \
+    -k $FIREWORKS_API_KEY -m accounts/fireworks/models/llama-v3-8b-instruct --token $HF_WRITE_TOKEN
+
+
+# Step 2: TTS the turn before last: audio input for the model
+
+just ds_tool tts -d fixie-ai/soda-audio -u fixie-ai/soda-audio -c "dialogue[-2]" -a audio_one_but_last -i eleven -V random --token $HF_WRITE_TOKEN -N 8
```
juberti, line 24: So this adds `alt_last_turn` as an assistant completion, and also `audio_one_but_last` as a TTS for the last user message. `audio_one_but_last` is a bit inscrutable to me, maybe `audio_last_user_message`? (and maybe `alt_last_assistant_message`?)

farzadab: I also don't like the name. The issue is there is no notion of user or assistant here, so those names would just be confusing.

ultravox/tools/wrappers.py
farzadab: Done.

scripts/dataset_creation/soda_audio.sh
juberti: hmm, agreed. `alt_last_turn` and `audio_second_last_turn` perhaps?

farzadab: Okay I'll change it. I'll have to reupload the dataset, so I'll block the PRs for that to land first.

PR #35, opened 2024-06-21
SODA Dataset for Training

farzadab:


ultravox/tools/infer_tool.py
```diff
@@ -91,7 +91,7 @@ def run_tui(
         print(f"--- Sample {index} ---")
     messages = sample.messages
     transcript = f' ["{sample.audio_transcript}"]' if sample.audio_transcript else ""
-    print(f"Q: {messages[0]['content']}{transcript}")
+    print(f"Q: {messages[-2]['content']}{transcript}")  # TODO: change?
```
juberti, line 94: what's the issue here?

ultravox/tools/infer_api.py
```diff
@@ -69,6 +69,7 @@ def infer_stream(
                     )
 
     def _build_message(self, sample: datasets.VoiceSample):
+        # TODO: fix for multiple messages?
```
juberti, line 72: yeah, we should probably return a list of messages here

ultravox/evaluation/eval_types.py
```diff
@@ -9,6 +9,7 @@ class Sample(dataclasses_json.DataClassJsonMixin):
     question: str
     generated_answer: str
     expected_answer: str
+    history: List[str] = dataclasses.field(default_factory=list)
```
juberti, line 12: Is this just alternating `content` strings? Might be worth adding a `Message` class with `role` to avoid accidents

ultravox/evaluation/gpt_eval.py
```diff
@@ -18,9 +19,9 @@
 """
 INSTRUCT_USER_PROMPT = """
 Using the supplied correct answer as an example, evaluate the model's ability to follow the instructions in the question below:
-Question: {question}
-Model answer: {generated_answer}
-Correct answer: {expected_answer}
+Question: {{ question }}
```
juberti, line 22: should we move these out to .jinja files too?

ultravox/tools/infer_tool.py
farzadab: This one is probably okay. I left the TODOs mainly so that I can come back and verify that.

ultravox/evaluation/eval_types.py
farzadab: Yeah I was thinking about that too.

Slightly orthogonal issue: for SODA, I didn't want to keep the "system" message at the top, but in some cases we might. We can fix this when we get to it.

ultravox/evaluation/gpt_eval.py
farzadab: Hmm, possible.

Not a strong opinion, but I'd rather have the system and user messages in the same place.
We can do that with Jinja, but that might get complicated again.

juberti: yeah, makes sense. maybe just split each eval and its prompts out into its own .py for now

ultravox/evaluation/eval_types.py
farzadab: Done.

ultravox/evaluation/gpt_eval.py
farzadab: Done. Separated evals and added a test for conversation eval because the Jinja template was getting messy.

farzadab:
The PR is ready.

ultravox/evaluation/gpt_eval_boolq.py
```diff
@@ -0,0 +1,23 @@
...
+BOOLQ_USER_PROMPT = """
+Using the supplied correct answer as ground truth, evaluate the model's answer to the question below:
+Question: {{ question }}
+Model answer: {{ generated_answer }}
+Correct answer: {{ expected_answer }}
+"""
+
+
+def evaluate_answer_boolq(sample: eval_types.Sample) -> eval_types.InstructResult:
+    return gpt_eval._evaluate_answer_gpt(BOOLQ_SYSTEM_PROMPT, BOOLQ_USER_PROMPT, sample)
```
juberti, line 23: probably want to remove the _ for this function, right?

ultravox/data/datasets.py
```diff
@@ -681,6 +683,51 @@ def _get_sample(self, idx, row) -> VoiceSample:
...
+        )
+        self._init_dataset(dataset)
+
+    def _get_sample(self, idx, row) -> VoiceSample:
+        turns = row["dialogue"]
+        # Make sure the last turn is the assistant's
+        roles = ["user", "assistant"] if len(turns) % 2 == 0 else ["assistant", "user"]
+
+        num_prompts = min(self._args.num_prompts, len(self.SYS_PROMPTS))
+        sys_prompt = self.SYS_PROMPTS[idx % num_prompts]
```
juberti, line 714: did we end up using a RNG for this sort of thing rather than the index?

juberti (approved):

ultravox/evaluation/eval_types.py
juberti: Did you decide to keep these messages as strings?

ultravox/data/datasets.py
juberti: (I forget where but we discussed adding a private RNG to datasets to allow them to simply pull a value from the RNG rather than using the index counter and various moduli)

ultravox/evaluation/eval_types.py
farzadab: You must be looking at an old version of the file:
image

ultravox/evaluation/gpt_eval_boolq.py
farzadab: Okay will do. Honestly the `_` in Python is weird to me since there's not much point to it (other than simply signalling to user about the developer's expectations).

ultravox/data/datasets.py
farzadab: Yes, the idea was that we do that in the next PR.

farzadab: Might as well just do it now I guess since I have the code.

farzadab: Done.

ultravox/evaluation/gpt_eval_boolq.py
farzadab: Done.

ultravox/data/datasets.py
```diff
@@ -258,14 +261,18 @@ def _load_audio_dataset(
         else:
             dataset = datasets.load_dataset(
                 path, name, split=split, trust_remote_code=True, streaming=streaming
-            ).cast_column("audio", datasets.Audio(sampling_rate=SAMPLE_RATE))
+            )
+            for column_name in self.BASE_AUDIO_COLUMNS:
+                dataset = dataset.cast_column(
+                    column_name, datasets.Audio(sampling_rate=SAMPLE_RATE)
+                )
```
farzadab, line 268: Bugfix for datasets that have audio column that is not named `audio`.

farzadab: This was not an issue for SODA since it was constructed with 16K Hz, but it was sloppy of me.

PR #36, opened 2024-06-23
Minor fixes to ds_tool and infer_tool

juberti:
- --upload_split param to allow the dest split to be different than the src split
- allow @file syntax for --prompt
- add retries and timeouts to TTS requests

### DIFF
--- a/ultravox/tools/ds_tool.py
+++ b/ultravox/tools/ds_tool.py
@@ -98,6 +98,7 @@ class DatasetToolArgs:
     num_workers: int = simple_parsing.field(default=16, alias="-w")
 
     upload_name: Optional[str] = simple_parsing.field(default=None, alias="-u")
+    upload_split: Optional[str] = simple_parsing.field(default=None)
     upload_branch: Optional[str] = simple_parsing.field(default="main", alias="-B")
     num_shards: Optional[int] = simple_parsing.field(default=None, alias="-N")
     private: bool = simple_parsing.field(default=False)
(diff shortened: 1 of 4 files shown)


ultravox/tools/ds_tool.py
```diff
@@ -132,6 +138,7 @@ def main(args: DatasetToolArgs):
         "token": token,
         "revision": args.upload_branch,
         "private": args.private,
+        "split": args.upload_split,
```
farzadab, line 141: Just noting that this is going to conflict with my PR, but it's okay to merge yours first.

farzadab (approved):

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -150,6 +159,7 @@ def main(args: DatasetToolArgs):
         "token": token,
         "revision": args.upload_branch,
         "private": args.private,
+        "split": args.upload_split,
```
farzadab, line 162: Wait, you didn't fix the conflict here!
Now on line 170 you'll be sending the `split` argument twice, and with different values!

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -126,6 +130,11 @@ class DatasetToolArgs:
         positional=True,
     )
 
+    def __post_init__(self):
+        assert (
+            not self.upload_split or self.dataset_split
+        ), "Must specify dataset_split when using upload_split"
```
farzadab, line 136: Sorry I didn't catch this the first time, but why can't `upload_split` be equal to `dataset_split` by default if not specified?

PR #47, opened 2024-06-26
Add adapter for HiSanta data

mdepinet:


mdepinet:
@farzadab It wasn't clear to me whether VoiceDatasetArgs are optional customizations to be used by some datasets or whether there are some that Datasets are required to respect.  (I imagine at least max_audio_duration_secs is required?)  Should be pretty easy to add support for the required ones once I know which those are.

Note to self: Need to set up new service account.

farzadab:
I believe `include_audio`, `shuffle`, `max_audio_duration_secs`, and `split` should be respected. The other args can be situational.

ultravox/data/datasets.py
```diff
@@ -681,6 +684,80 @@ def _get_sample(self, idx, row) -> VoiceSample:
...
+    def __init__(self, args: VoiceDatasetArgs, subset: Subset = Subset.BEST) -> None:
+        super().__init__()
+        self._args = args  # TODO(mdepinet): Respect whatever args we need to.
+        self._bucket = gcs.Client().get_bucket("hisanta-dataset")
+
+        self._conversations: list[str] = json.loads(
+            self._bucket.get_blob(f"{subset}.json").download_as_bytes()
+        )["conversations"]
+        """List of references to conversation metadata JSON files in the bucket.
+        These all look like {conversation_id}/metadata.json."""
```
farzadab, line 706: Nit: Why use strings for comments?

ultravox/data/datasets.py
```diff
@@ -681,6 +684,80 @@ def _get_sample(self, idx, row) -> VoiceSample:
...
+        history = [{"role": "system", "content": system_prompt}]
+        for message in conversation["messages"]:
+            if message["role"] == "assistant":
+                history.append({"role": "assistant", "content": message["message"]})
+            else:
+                audio = self._bucket.get_blob(
+                    f"{conversation_id}/{message['speech']}"
+                ).download_as_bytes()
+                yield VoiceSample(
+                    messages=[*history, {"role": "user", "content": "<|audio|>"}],
```
farzadab, line 742: Current assumption is that the last message should be the assistant message.

ultravox/data/datasets.py
```diff
@@ -681,6 +684,80 @@ def _get_sample(self, idx, row) -> VoiceSample:
...
+        """List of system prompts for each agent, with the time at which each became effective.
+        These all look like {"agent_id": AGENT_ID, "prompts": [{"start": ISO_TIMESTAMP, "prompt": PROMPT}]}
+        where the prompts are in chronological order."""
+
+    def __iter__(self) -> Generator[VoiceSample, Any, None]:
+        worker_info = data.get_worker_info()
+        start = worker_info.id if worker_info else 0
+        increment = worker_info.num_workers if worker_info else 1
+        for i in range(start, len(self._conversations), increment):
+            yield from self._from_conversation(i)
```
farzadab, line 720: We'll probably have to experiment with how to form our samples here.
There are multiple issues to consider:
1. How to do shuffle
2. The length of each sample should be regulated: `max_audio_duration_secs` was an attempt at this, but generally the bottleneck is GPU memory

mdepinet:
Putting this on ice for now.

PR #48, opened 2024-07-09
More validations

farzadab:
This PR separates validation data_sets from training data_sets and adds separate evaluations for the following datasets both in the audio and text-only modes: `["heysquad_human", "anyinstruct", "soda", "peoplespeech"]`

Example perplexity/loss curves:
image

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -218,7 +218,7 @@ class VoiceDatasetArgs:
 
 
 def _get_messages(
-    *turns: List[str], sys_prompt: Optional[str] = None, assistant_last: bool = True
+    *turns: str, sys_prompt: Optional[str] = None, assistant_last: bool = True
 ) -> List[Dict[str, str]]:
     """
     Convert a list of strings into a list of messages, alternating between user and assistant.
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -29,8 +29,8 @@ ANSWER_TASK = "answer"
 
 TRANSCRIBE_PROMPTS = [
     # from Gazelle
-    "Transcribe <|audio|>",
-    "Transcribe exactly what is said here <|audio|>",
+    "Transcribe\n<|audio|>",
+    "Transcribe exactly what is said here\n<|audio|>",
     "Repeat exactly what is written here: <|audio|>",
     "Write exactly what was said: <|audio|>",
     "First listen to the clip. Then, transcribe exactly what is said. <|audio|>",
@@ -217,6 +217,30 @@ class VoiceDatasetArgs:
             self.split = DatasetSplit(self.split.lower())
 
 
+def _get_messages(
+    *turns: List[str], sys_prompt: Optional[str] = None, assistant_last: bool = True
+) -> List[Dict[str, str]]:
+    """
+    Convert a list of strings into a list of messages, alternating between user and assistant.
+    If `sys_prompt` is set, it is prepended as a system message.
+    If `assistant_last` is True, the assistant's message is the last one.
+    """
+    messages = []
+
+    if sys_prompt:
+        messages.append({"role": "system", "content": sys_prompt})
+
+    roles = ["user", "assistant"]
+
+    # Make sure the last turn is the assistant's iff assistant_last is True.
+    if (len(turns) + assistant_last) % 2 == 0:
+        roles = roles[::-1]
+
+    messages += [{"role": roles[i % 2], "content": c} for i, c in enumerate(turns)]
+
+    return messages
+
+
 class VoiceDataset(abc.ABC, data.IterableDataset):
     """
     Base class for streaming voice datasets.
@@ -312,19 +336,13 @@ class VoiceDataset(abc.ABC, data.IterableDataset):
     ) -> List[Dict[str, str]]:
         prompt = self._get_answer_prompt() if self._args.include_audio else question
         user_content = f"{context}\n\n{prompt}" if context else prompt
-        return [
-            {"role": "user", "content": user_content},
-            {"role": "assistant", "content": answer},
-        ]
+        return _get_messages(user_content, answer)
 
     def _get_transcribe_messages(self, text: str) -> List[Dict[str, str]]:
         prompt = self._get_transcribe_prompt()
         if not self._args.include_audio:
             prompt = prompt.replace("<|audio|>", text)
-        return [
-            {"role": "user", "content": prompt},
-            {"role": "assistant", "content": text},
-        ]
+        return _get_messages(prompt, text)
 
     def _get_audio(
         self, row: transformers.BatchFeature, column_name: str = "audio"
@@ -527,6 +545,8 @@ class QAVoiceDatasetMixin(VoiceDataset):
     # In most cases there is no extra prompt-suffix needed
     PROMPT_SUFFIXES = [""]
 
+    # TODO: combine `_get_query_prompt` and `_get_answer_messages` into a single method
+    # and use this mixin for all non-ASR datasets.
     def _get_query_prompt(self, question_str: str, context: str) -> str:
         """
         Creates a random prompt for a QA sample with a passage and question.
@@ -592,13 +612,10 @@ class BoolQWithExtendedAnswerDataset(BoolQDataset, QAVoiceDatasetMixin):
         user_message = self._get_query_prompt(
             question_str=row["question"], context=row["passage"]
         )
-        messages = [
-            {"role": "user", "content": user_message},
-            {
-                "role": "assistant",
-                "content": f"{row['explanation']}\n{answer_prompt}{answer}",
-            },
-        ]
+
+        messages = _get_messages(
+            user_message, f"{row['explanation']}\n{answer_prompt}{answer}"
+        )
 
         return self._make_sample(
             messages, self._get_audio(row), audio_transcript=row["question"]
@@ -638,10 +655,7 @@ class HeySQuADHumanDataset(QAVoiceDatasetMixin):
         prompt = self._get_query_prompt(
             question_str=row["question"], context=row["context"]
         )
-        messages = [
-            {"role": "user", "content": prompt},
-            {"role": "assistant", "content": row["answers"][0]["text"]},
-        ]
+        messages = _get_messages(prompt, row["answers"][0]["text"])
         return self._make_sample(
             messages, self._get_audio(row), audio_transcript=row["question"]
         )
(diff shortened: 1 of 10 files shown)


farzadab:
For some reason I had assumed that I had merged this PR but it was just sitting there for a week!

ultravox/data/datasets.py
```diff
@@ -308,8 +318,11 @@ def _get_answer_messages(
         ]
 
     def _get_transcribe_messages(self, text: str) -> List[Dict[str, str]]:
+        prompt = self._get_transcribe_prompt()
+        if not self._args.include_audio:
+            prompt = prompt.replace("<|audio|>", text)
```
juberti, line 323: Maybe put quotes around the text content so it ends up being like `Transcribe "Hello world"`

ultravox/data/datasets.py
```diff
@@ -492,78 +505,95 @@ def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
...
         "",
         "The answer is: ",
         "Result: ",
         "Conclusion: ",
     ]
+    # In most cases there is no extra prompt-suffix needed
+    PROMPT_SUFFIXES = [""]
 
-    def _get_query_prompt(self) -> str:
+    def _get_query_prompt(self, question_str: str, context: str) -> str:
```
juberti, line 530: this ends up overlapping with the functionality in `_get_answer_messages`. Should we remove `_get_answer_prompt` and `_get_answer_messages` and just use this mixin for all non-transcribe datasets?

ultravox/data/datasets.py
```diff
@@ -492,78 +505,95 @@ def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
...
 
-    def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
         answer = "True" if row["answer"] else "False"
-        answer_prompt = self._choice(self.ANSWER_PROMPTS)
-        query_prompt = self._get_query_prompt()
-        user_content = query_prompt.format(
-            question="<|audio|>" if self._args.include_audio else row["question"],
-            context=row["passage"],
+        answer_prompt = self._choice(self.ANSWER_PREFIX)
+        user_message = self._get_query_prompt(
```
juberti, line 592: suggest moving the context length check to `_get_query_prompt` so that you don't need to do this check in every dataset class

ultravox/data/datasets.py
```diff
@@ -492,78 +505,95 @@ def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
...
+            question_str=row["question"], context=row["passage"]
         )
         messages = [
-            {"role": "user", "content": user_content},
+            {"role": "user", "content": user_message},
```
juberti, line 596: suggest adding a builder method like `_get_messages(user, assistant)` so you don't need to build this message array by hand in each class.

ultravox/data/datasets.py
```diff
@@ -575,6 +605,99 @@ def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
...
+        Example conversation
+            <|user|> Context: {context}
+            Question: {question}
+            <|assistant|> {answer}
+        """
+        if len(row["raw_document_text"]) > self._args.max_context_length:
+            # Skip samples with long context
+            return None
+
+        print(row["question_id"], f"len spans: {len(row['answer_spans']['answer'])}")
```
juberti, line 685: remove?

ultravox/data/datasets.py
```diff
@@ -575,6 +605,99 @@ def _get_sample(self, row: transformers.BatchFeature) -> VoiceSample:
...
+    The following datasets are supposed to be included, but I haven't found them everywhere:
+      * WebQuestions (for questions whose question_id starts with 'wq-')
+      * CuratedTREC (for questions whose question_id starts with 'trec-')
+      * Spoken Wikipedia
+
+
+    Splits: train, validation, test, verified_test
+    """
+
+    BASE_AUDIO_COLUMNS = ["question_audio", "document_audio"]
```
juberti, line 665: is this used?

ultravox/training/train.py
```diff
@@ -179,29 +184,39 @@ def main() -> None:
...
         )
+        val_ds_args_text = copy.copy(val_ds_args)
+        val_ds_args_text.include_audio = False
+        val_datasets = {
+            k: prepare_dataset(
+                dataset_names=val_sets[k],
+                train_on_inputs=args.train_on_inputs,
+                repeat_data=args.repeat_data,
+                processor=processor,
+                num_samples=(
```
juberti, line 203: I didn't follow what's happening with "matchtrain". Can you add a comment to explain?

ultravox/training/train.py
```diff
@@ -211,7 +226,7 @@ def main() -> None:
     trainer = transformers.Seq2SeqTrainer(
         model,
         train_dataset=train_dataset,
-        eval_dataset=val_dataset,
+        eval_dataset=val_datasets,
```
juberti, line 229: does this work OK as a collection, or do we need to wrap this dict of datasets with a Combine or Interleave wrapper?

ultravox/data/datasets.py
farzadab: Hmm, makes sense.
I'm wondering if we need this for audio as well then. Assuming perfect alignment, audio and text should be replaceable.

ultravox/data/datasets.py
farzadab: Do you mean is `BASE_AUDIO_COLUMNS` ever used or are you asking whether we need to convert `document_audio` if we never use it?

image

ultravox/training/train.py
farzadab: Will do. I was struggling with the name too. Let's disambiguate here first.

The idea is that classically both training and validation loss come from the same data distribution. Having that still has some advantages, simplest one being a way to show overfitting.

We have broken that assumption with the "dataset specific validations", but I still wanted to have the original. I wasn't sure how to name this. Since the point was to "match the training dataset distribution", I called it "matchtrain", but I'm open to ideas.

ultravox/training/train.py
farzadab: Yes it works. If you give it a dict, then all of them are run separately and you get their metrics in W&B separately as well.

That's why if you look at the image at the top of the PR, you can see curves for: "peoplespeech", "text_peoplespeech", "soda", "text_soda", and so on.

ultravox/data/datasets.py
farzadab: @zqhuang211 do you have any strong opinions here? I.e., what is the right text version of this audio prompt?

Choices:
1. `Transcribe <|audio|>` ->  `Transcribe Hello world`
1. `Transcribe <|audio|>` ->  `Transcribe "Hello world"`
1. `Transcribe "<|audio|>"` ->  `Transcribe "Hello world"`

Or even:
1. `Transcribe\n<|audio|>` ->  `Transcribe\nHello world`

ultravox/data/datasets.py
farzadab: I was being a bit wasteful by keeping `document_audio` at the cost of making sure we don't forget (e.g. if a child-class wants to use it). Didn't have a strong opinion here though.

ultravox/data/datasets.py
farzadab: Okay so the main issue I have for combining them is that `ANSWER_PROMPTS` don't assume full audio-text equivalence. For example, we have the prompt `Listen to <|audio|> and respond to it`.

Now, what is the text-equivalent of that exactly? In `_get_answer_messages` we're saying the whole prompt (with the audio in it) is equivalent to the question's text. Maybe that's okay there (I've never liked that), but when trying to combine them things get a bit weird. Also, what is the role of `self._args.prompt` at that point? Does it go at the start or at the end?

I propose to get rid of `ANSWER_PROMPTS` and essentially only have `<|audio|>` there. I will instead embellish `QUERY_PREFIX = ["Question: ", "Question:\n", "Q: ", "Q:\n", "Query: ", "Query:\n"]` with variations like `ANSWER_PROMPT`, but make them agnostic to the modality of the input. Thoughts?

farzadab: Since this won't be easily verifiable, the prudent thing to do is to train with it and test the resulting model. That would probably require a new PR though.

juberti: I think if we want multi-task behavior, we need text instructions. But I agree we should make the result agnostic of modality as you describe. For now just maybe leave a comment with the plan here and we can revisit it in a future PR.

ultravox/data/datasets.py
juberti: I lean toward 4) as perhaps the cleanest approach.

ultravox/data/datasets.py
juberti: OK, I didn't realize this was overriding a base class constant. A little confusing perhaps but we can worry about this later.

ultravox/data/datasets.py
farzadab: Switching all `Transcribe <|audio|>` to `Transcribe\n<|audio|>`.
I didn't switch "all" prompts since I felt it was good to have them in training still.

ultravox/data/datasets.py
farzadab: Done.

ultravox/training/train.py
farzadab: Done.

ultravox/data/datasets.py
farzadab: Done.

farzadab:
PTAL @juberti. All comments were applied.

ultravox/training/train.py
```diff
@@ -179,29 +187,39 @@ def main() -> None:
...
+        val_datasets = {
+            k: prepare_dataset(
+                dataset_names=val_sets[k],
+                train_on_inputs=args.train_on_inputs,
+                repeat_data=args.repeat_data,
+                processor=processor,
+                num_samples=(
+                    args.val_num_samples
+                    if k == "matchtrain"
+                    else args.val_num_samples // 2
```
juberti, line 209: Can you explain the // 2? Not quite sure why we're doing this.

juberti (approved):

ultravox/training/train.py
farzadab: It's just to make things faster, since the `validation` now takes 9x time now.
But I agree there's not a huge reason for separating `matchtrain`. I'll reduce `val_num_samples` for all datasets instead.

farzadab: Removed and reduced `val_num_samples` everywhere by 2.

PR #52, opened 2024-07-19
Update ds_tool.py

zqhuang211:
This PR addresses several issues in ds_tool.py and provides better support for creating synthesized speech/text

- Simplified the logic for consistent naming of subsets/splits for datasets (both input and output).
- Fixed bug related to ntlk/multiprocessing on Mac.
- Added option to format text fields before generation.
- Added option to control memory consumption per process.
- Added template for creating continuations.

### DIFF
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -7,6 +7,8 @@ import datasets
 import jinja2
 import openai
 import simple_parsing
+from jinja2 import StrictUndefined
+from jinja2 import TemplateError
 
 from ultravox.data.text_proc import format_asr_text
 from ultravox.tools.ds_tool import caching
@@ -51,6 +53,17 @@ class TtsTask:
         # using a Jinja template for some added flexibility
         # The {{ var }} syntax is how Jinja denotes variables
         text = jinja2.Template("{{" + self.column_name + "}}").render(**sample)
+
+        try:
+            text = jinja2.Template("{{" + self.column_name + "}}").render(**sample)
+        except TemplateError as e:
+            print(f"Error rendering template: {e}")
+            print(f"column_name: {self.column_name}")
+            print(f"sample keys: {list(sample.keys())}")
+            raise ValueError(
+                f"Template rendering failed. Make sure column_name exists in the sample."
+            ) from e
+
         text = text["text"] if isinstance(text, dict) else text
         sample[self.audio_column_name] = tts_client.tts(text, self.voice)
         return sample
@@ -91,7 +104,18 @@ class TextGenerationTask:
     def _map_sample(self, sample):
         for field in self.format_fields:
             sample[field] = format_asr_text(sample[field])
-        rendered = jinja2.Template(self.template).render(**sample, json_dump=json.dumps)
+
+        try:
+            rendered = jinja2.Template(self.template, undefined=StrictUndefined).render(
+                **sample, json_dump=json.dumps
+            )
+        except TemplateError as e:
+            print(f"Error rendering template: {e}")
+            print(f"template: {self.template}")
+            print(f"sample keys: {list(sample.keys())}")
+            raise ValueError(
+                f"Template rendering failed. Make sure all keys in the template exist in the sample."
+            ) from e
 
         if self.json_mode:
             turns = json.loads(rendered)
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -169,7 +169,7 @@ class DatasetToolArgs:
     )
 
     def __post_init__(self):
-        if not self.upload_subset:
+        if not self.upload_subset and self.dataset_subset:
             self.upload_subset = self.dataset_subset
         if self.dataset_split and not self.upload_split:
             self.upload_split = self.dataset_split
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -26,7 +26,6 @@ class TtsTask:
     audio_column_name: Optional[str] = simple_parsing.field(default=None, alias="-a")
     voice: Optional[str] = simple_parsing.field(default=None, alias="-V")
     sample_rate: int = simple_parsing.field(default=16000, alias="-r")
-    write_batch_size: int = 1000
     format_fields: List[str] = simple_parsing.field(default_factory=list)
 
     def __post_init__(self):
@@ -39,10 +38,12 @@ class TtsTask:
             provider=self.implementation,
         )
 
-    def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Dataset:
+    def map_split(
+        self, ds_split: datasets.Dataset, num_proc: int, writer_batch_size: int
+    ) -> datasets.Dataset:
         print(f'TTS mapping "{self.column_name}" to "{self.audio_column_name}"...')
         return ds_split.map(
-            self._map_sample, num_proc=num_proc, writer_batch_size=self.write_batch_size
+            self._map_sample, num_proc=num_proc, writer_batch_size=writer_batch_size
         ).cast_column(
             self.audio_column_name, datasets.Audio(sampling_rate=self.sample_rate)
         )
@@ -80,7 +81,6 @@ class TextGenerationTask:
     api_key: Optional[str] = simple_parsing.field(default=None, alias="-k")
     max_tokens: int = 128
     temperature: float = 0
-    write_batch_size: int = 1000
     format_fields: List[str] = simple_parsing.field(default_factory=list)
 
     def __post_init__(self):
@@ -95,10 +95,12 @@ class TextGenerationTask:
             with open(self.template[1:], "r") as template_file:
                 self.template = template_file.read()
 
-    def map_split(self, ds_split: datasets.Dataset, num_proc: int) -> datasets.Dataset:
+    def map_split(
+        self, ds_split: datasets.Dataset, num_proc: int, writer_batch_size: int
+    ) -> datasets.Dataset:
         print(f'Generating "{self.new_column_name}" with template:\n{self.template}')
         return ds_split.map(
-            self._map_sample, num_proc=num_proc, writer_batch_size=self.write_batch_size
+            self._map_sample, num_proc=num_proc, writer_batch_size=writer_batch_size
         )
 
     def _map_sample(self, sample):
@@ -142,7 +144,7 @@ class TextGenerationTask:
 #   just ds_tool textgen --new_column_name continuation --dataset_name openslr/librispeech_asr --dataset_subset clean --dataset_split train.360 \
 #        --shuffle --format_fields text --upload_name fixie-ai/librispeech_asr --private --base_url https://api.fireworks.ai/inference/v1 \
 #        --api_key $FIREWORKS_API_KEY --token $HF_TOKEN --language_model accounts/fireworks/models/llama-v3-8b-instruct \
-#        --template @ultravox/tools/ds_tool/continuation.jinja --max_tokens 64 --num_workers 30 --write_batch_size 30
+#        --template @ultravox/tools/ds_tool/continuation.jinja --max_tokens 64 --num_workers 30 --writer_batch_size 30
 @dataclasses.dataclass
 class DatasetToolArgs:
     # HF source dataset parameters
@@ -155,6 +157,7 @@ class DatasetToolArgs:
     shuffle_seed: int = simple_parsing.field(default=42)
     num_samples: Optional[int] = simple_parsing.field(default=None, alias="-n")
     num_workers: int = simple_parsing.field(default=16, alias="-w")
+    writer_batch_size: int = simple_parsing.field(default=1000)
 
     # HF destination dataset parameters
     upload_name: Optional[str] = simple_parsing.field(default=None, alias="-u")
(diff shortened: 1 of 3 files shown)


zqhuang211:
@farzadab check failed due to the hack solution for the nltk bug, borrowed from:  https://github.com/nltk/nltk/issues/2949

ultravox/data/text_proc.py:4: error: Incompatible types in assignment (expression has type "None", target has type Module)  [assignment]

`
sys.modules["tkinter"] = None
`

Any idea how to fix/ignore the check failure?

farzadab:
To ignore you can add a `# type: ignore` comment on the same line.
I haven't looked closely to know why the check is failing though.

zqhuang211:
updated to ignore the error:
`
sys.modules["tkinter"] = None  # type: ignore
`

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -23,6 +26,8 @@ class TtsTask:
     audio_column_name: Optional[str] = simple_parsing.field(default=None, alias="-a")
     voice: Optional[str] = simple_parsing.field(default=None, alias="-V")
     sample_rate: int = simple_parsing.field(default=16000, alias="-r")
+    write_batch_size: int = 1000
```
juberti, line 29: suggest adding this as an argument to `DatasetToolArgs` since it's common across tasks, similar to `num_workers`.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -36,14 +41,29 @@ def __post_init__(self):
...
+        ).cast_column(
             self.audio_column_name, datasets.Audio(sampling_rate=self.sample_rate)
         )
 
     def _map_sample(self, sample):
+        for field in self.format_fields:
+            sample[field] = format_asr_text(sample[field])
         # using a Jinja template for some added flexibility
         # The {{ var }} syntax is how Jinja denotes variables
         text = jinja2.Template("{{" + self.column_name + "}}").render(**sample)
```
juberti, line 55: is this line still needed?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -1,13 +1,16 @@
...
+from typing import Any, Dict, List, Optional, Union
 
 import datasets
 import jinja2
 import openai
 import simple_parsing
+from jinja2 import StrictUndefined
+from jinja2 import TemplateError
 
+from ultravox.data.text_proc import format_asr_text
```
juberti, line 13: per Google style, this should just import `text_proc` and then you can do `text_proc.format_asr_text` below.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -75,10 +97,25 @@ def __post_init__(self):
...
         print(f'Generating "{self.new_column_name}" with template:\n{self.template}')
-        return ds_split.map(self._map_sample, num_proc=num_proc)
+        return ds_split.map(
+            self._map_sample, num_proc=num_proc, writer_batch_size=self.write_batch_size
+        )
 
     def _map_sample(self, sample):
-        rendered = jinja2.Template(self.template).render(**sample, json_dump=json.dumps)
+        for field in self.format_fields:
+            sample[field] = format_asr_text(sample[field])
```
juberti, line 106: Could this formatting be done via the jinja template? It does seem kind of specific to this one particular type of completion.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -131,55 +173,57 @@ class DatasetToolArgs:
     )
 
     def __post_init__(self):
-        assert (
-            not self.upload_split or self.dataset_split
-        ), "Must specify dataset_split when using upload_split"
+        assert self.dataset_subset, "dataset_subset must be specified"
```
juberti, line 176: I don't think this condition is always true?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -131,55 +173,57 @@ class DatasetToolArgs:
...
 
-    token = args.token or os.environ.get("HF_TOKEN")
     hub_args: Dict[str, Any] = {
-        "config_name": args.dataset_subset or "default",
-        "token": token,
+        "config_name": args.upload_subset,
+        "token": args.token or os.environ.get("HF_TOKEN"),
         "revision": args.upload_branch,
         "private": args.private,
-        "split": args.upload_split,
```
juberti, line 162: I think this removes the ability to just update a single split without touching other splits in a dataset.

zqhuang211: The specified subset/config of the dataset is processed one split a time, and the push_to_hub call always includes the split name (original or new) in its arguments:

ds_split.push_to_hub(args.upload_name, split=upload_split, **hub_args)

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: you are right.

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: I think this is consistent with the selective use of text_proc.format_asr_text in ultravox.data.datasets.py for certain datasets.

ultravox/tools/ds_tool/ds_tool.py
farzadab: I'm still confused. Why do you do the same thing twice?
image

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -131,55 +176,58 @@ class DatasetToolArgs:
...
-        else:
-            data_dict.push_to_hub(args.upload_name, **hub_args)
-    except Exception as e:
-        print(f"Failed to push to hub: {e}")
-
-        # If the push fails or upload_name is not specified, save the data locally.
-        for split in data_dict.keys():
-            output_name = f"{split}-00000-of-00001.parquet"
-            data_dict[split].to_parquet(output_name)
+    for split, ds_split in ds.items():
```
farzadab, line 207: The sequential upload doesn't work if you change the data format (e.g. add a column). That's why we kept both options.

farzadab: You can create a new dataset to push to and rename it later, but that will completely destroy the history.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -34,16 +38,33 @@ def __post_init__(self):
...
-        return ds_split.map(self._map_sample, num_proc=num_proc).cast_column(
+        return ds_split.map(
+            self._map_sample, num_proc=num_proc, writer_batch_size=writer_batch_size
+        ).cast_column(
             self.audio_column_name, datasets.Audio(sampling_rate=self.sample_rate)
         )
 
     def _map_sample(self, sample):
+        for field in self.format_fields:
+            sample[field] = text_proc.format_asr_text(sample[field])
```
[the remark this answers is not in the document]
zqhuang211: Now I see what you and Justin meant by this, something like {{ text | format_asr_text }}, right?

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: good catch, this is an error.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -114,12 +157,14 @@ class DatasetToolArgs:
     shuffle_seed: int = simple_parsing.field(default=42)
     num_samples: Optional[int] = simple_parsing.field(default=None, alias="-n")
     num_workers: int = simple_parsing.field(default=16, alias="-w")
+    writer_batch_size: int = simple_parsing.field(default=1000)
```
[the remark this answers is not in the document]
zqhuang211: Yes, we can run out of memory. Each worker caches write_batch_size amount of samples during map processing before writing out to disk. The memory consumption is a linear function of both num_workers and writer_batch_size.

ultravox/tools/ds_tool/ds_tool.py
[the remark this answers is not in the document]
farzadab: Yeah. The "filter" format that you used needs [to be added under `environment.filters`](https://jinja.palletsprojects.com/en/3.1.x/api/#custom-filters)

You can even use it as a function: `{{ format_asr_text(text) }}`. For this I think you just need to pass the function as another variable.

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: This formatting is now moved to the jinja template.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -23,6 +26,7 @@ class TtsTask:
     audio_column_name: Optional[str] = simple_parsing.field(default=None, alias="-a")
     voice: Optional[str] = simple_parsing.field(default=None, alias="-V")
     sample_rate: int = simple_parsing.field(default=16000, alias="-r")
+    format_fields: List[str] = simple_parsing.field(default_factory=list)
```
farzadab, line 29: not required anymore I believe.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -34,16 +38,29 @@ def __post_init__(self):
...
+            text = jinja2.Template("{{" + self.column_name + "}}").render(**sample)
+        except TemplateError as e:
+            print(f"Error rendering template: {e}")
+            print(f"column_name: {self.column_name}")
+            print(f"sample keys: {list(sample.keys())}")
+            raise ValueError(
+                f"Template rendering failed. Make sure column_name exists in the sample."
+            ) from e
+
         text = text["text"] if isinstance(text, dict) else text
```
farzadab, line 64: nit and not related to this PR: it feels like this line is also not needed anymore since it can/should be folded into the `column_name` string (cc @juberti).

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -60,6 +77,7 @@ class TextGenerationTask:
     api_key: Optional[str] = simple_parsing.field(default=None, alias="-k")
     max_tokens: int = 128
     temperature: float = 0
+    format_fields: List[str] = simple_parsing.field(default_factory=list)
```
farzadab, line 80: not required anymore I believe.

farzadab (approved):

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: arghh, it was removed but somehow not saved properly before committing.

ultravox/tools/ds_tool/ds_tool.py
juberti: agreed, we might even in the future allow for user-supplied templates which would remove the need for `column_name`

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -131,55 +169,58 @@ class DatasetToolArgs:
...
+        ds = datasets.DatasetDict({args.upload_split: ds})
+
+    if len(ds) > 1 and args.upload_split:
+        raise ValueError("Cannot upload multiple splits to a single split")
 
-    token = args.token or os.environ.get("HF_TOKEN")
     hub_args: Dict[str, Any] = {
-        "config_name": args.dataset_subset or "default",
-        "token": token,
+        "config_name": args.upload_subset if args.upload_subset else "default",
```
juberti, line 192:
```suggestion
        "config_name": args.upload_subset or "default",
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -131,55 +169,58 @@ class DatasetToolArgs:
     )
 
     def __post_init__(self):
-        assert (
-            not self.upload_split or self.dataset_split
-        ), "Must specify dataset_split when using upload_split"
+        if not self.upload_subset:
```
juberti, line 172:
```suggestion
        if not self.upload_subset and self.dataset_subset:
```

ultravox/tools/ds_tool/ds_tool.py
juberti: I think this is still unresolved.

juberti (commented):
Looks good, but I think there's still an issue with the data-dict processing per the comments there.

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: Isn't the two essentially the same, except one is shorter?

juberti: perhaps; I was suggesting matching the code on lines 174-175, for easier readability.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -142,32 +180,40 @@ def main(args: DatasetToolArgs):
...
     }
-    if args.num_shards is not None:
-        hub_args["num_shards"] = {split: args.num_shards for split in data_dict.keys()}
 
+    if args.num_shards is not None:
+        hub_args["num_shards"] = args.num_shards
+    
     try:
         if args.dataset_split:
+            upload_split = args.upload_split or args.dataset_split
```
juberti, line 214: I think `upload_split` will already be set from `__post_init__`, right?

zqhuang211: good catch

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -142,32 +180,40 @@ def main(args: DatasetToolArgs):
...
-
-    token = args.token or os.environ.get("HF_TOKEN")
+        data_dict[split] = args.task.map_split(
+            ds_split, args.num_workers, args.writer_batch_size
+        )
+    
     hub_args: Dict[str, Any] = {
-        "config_name": args.dataset_subset or "default",
-        "token": token,
+        "config_name": args.upload_subset if args.upload_subset else "default",
```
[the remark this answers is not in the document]
zqhuang211: must be some issue with my workflow, it should have been fixed.

PR #53, opened 2024-07-23
CoVoST 2 dataset

farzadab:
Separating the PR to make it easier.

The CoVoST2 dataset can be used by specifying the subset name and no default was used since neither option seemed like a reasonable default.

Examples for setting the dataset as well as the name: `covost2:en_zh-CN` (English to Chinese), `covost:es_en` (Spanish to English).

Related to #50

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -848,11 +848,11 @@ class CoVoST2Dataset(VoiceDataset):
 
     # We currently don't use this dataset for training, so mainly the first prompt it ever used.
     TRANSLATE_PROMPTS = [
-        "Translate the following into {target} language: <|audio|>",
         "Translate the following into {target}: <|audio|>",
+        "Translate the following into {target} language: <|audio|>",
         "Please convert the following into {target}.\n<|audio|>",
         "Could you translate this to {target} language?\n<|audio|>",
-        "Translate the text below to {target}. <|audio|>",
+        "Translate the text below to {target}.\n<|audio|>",
         "Translate the subsequent text into {target} language. <|audio|>",
         "Can you translate this into the {target} language?\n<|audio|>",
         "Transform the following to {target}: <|audio|>",
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -875,7 +875,7 @@ class CoVoST2Dataset(VoiceDataset):
         transcript = row["sentence"]
         translation = row["translation"]
         if not self._args.include_audio:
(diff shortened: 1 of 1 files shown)


ultravox/data/datasets.py
```diff
@@ -805,6 +805,85 @@ def _get_sample(self, row) -> VoiceSample:
...
+        "id": "Indonesian",
+        "ar": "Arabic",
+        "ta": "Tamil",
+        "lv": "Latvian",
+        "ja": "Japanese",
+        "fr": "French",
+        "es": "Spanish",
+    }
+
+    # We currently don't use this dataset for training, so mainly the first prompt it ever used.
```
juberti, line 849: I didn't understand this comment.

juberti (approved):

ultravox/data/datasets.py
farzadab: I was just noting that since we only use CoVoST for validation/testing and since there we always pin `num_prompts` to 1, that means we never use the other prompts. I still kept them just in case though.

ultravox/data/datasets.py
```diff
@@ -805,6 +805,85 @@ def _get_sample(self, row) -> VoiceSample:
...
+    TRANSLATE_PROMPTS = [
+        "Translate the following into {target} language: <|audio|>",
+        "Translate the following into {target}: <|audio|>",
```
zqhuang211, line 852: This prompt is more concise and clear as the default prompt

> Translate the following into English: <|audio|>

farzadab: So switch prompt 1 and 2? Done.

ultravox/data/datasets.py
zqhuang211: Agree that we would only use CoVoST for validation/testing for now, but eventually we may also want to use it for training, similar to what's reported in the llama3 paper. 

by the way, some language pairs already have the training split included.

farzadab: Yeah we can always add them if need be. The code is attached in the issue and it's only a few lines of code.

Before doing that I wanted to explore and see if we can somehow reduce the amount of duplicated data, which at the current rate was a bit ridiculous (e.g. using a custom dataset loader). Large datasets are quite hard to manage as any small modification can require downloading and uploading huge amounts of data.

As for the included `train` sets, I kept the `XX to EN` sets since there's no duplication anyway.
For the `EN to XX` I kept only one (or two maybe?). All the other audio samples are duplicated.

PR #55, opened 2024-07-24
Make so infer_tools works with a single arg for filename

cdiddy77:
Previous to [this change](https://github.com/fixie-ai/ultravox/pull/35), it would be possible to say:

```
just infer -f path/to/file.wave
```
or 
```
just infer -f path/to/file.wav --prompt "some prompt"
```

But that broke. 

I have no idea if this is the right fix or not. Happy to run any local tests to verify.


juberti (approved):
Thanks for the fix, this looks right to me.

PR #56, opened 2024-07-25
Llama3.1

farzadab:
This PR enables Llama 3.1 training and loading.

Using Llama 3.1 required us to upgrade transformers. The new version changes caching so there are some changes related to that as well.

### DIFF
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -10,8 +10,7 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/$UV_CONFIG
+  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/llama3_whisper.yaml
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   UV_BRANCH: main
-  UV_CONFIG: asr_llama.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -10,7 +10,8 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/llama3_whisper.yaml
+  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   UV_BRANCH: main
+  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper.yaml
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -848,11 +848,11 @@ class CoVoST2Dataset(VoiceDataset):
 
     # We currently don't use this dataset for training, so mainly the first prompt it ever used.
     TRANSLATE_PROMPTS = [
-        "Translate the following into {target} language: <|audio|>",
         "Translate the following into {target}: <|audio|>",
+        "Translate the following into {target} language: <|audio|>",
         "Please convert the following into {target}.\n<|audio|>",
         "Could you translate this to {target} language?\n<|audio|>",
-        "Translate the text below to {target}. <|audio|>",
+        "Translate the text below to {target}.\n<|audio|>",
         "Translate the subsequent text into {target} language. <|audio|>",
         "Can you translate this into the {target} language?\n<|audio|>",
         "Transform the following to {target}: <|audio|>",
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -875,7 +875,7 @@ class CoVoST2Dataset(VoiceDataset):
         transcript = row["sentence"]
         translation = row["translation"]
         if not self._args.include_audio:
-            prompt.replace("<|audio|>", transcript)
+            prompt = prompt.replace("<|audio|>", transcript)
 
         return self._make_sample(
             _get_messages(prompt, translation),
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -29,8 +29,8 @@ ANSWER_TASK = "answer"
 
 TRANSCRIBE_PROMPTS = [
     # from Gazelle
-    "Transcribe <|audio|>",
-    "Transcribe exactly what is said here <|audio|>",
+    "Transcribe\n<|audio|>",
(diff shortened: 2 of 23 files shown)


mcloud.yaml
```diff
@@ -10,8 +10,7 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/$UV_CONFIG
+  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train --config_path ultravox/training/configs/llama3_whisper.yaml
```
farzadab, line 13: Since we can't control `UV_CONFIG` from the env variables, I'm not sure if it has any use at the moment.
I also changed the default to whisper.

ultravox/training/train.py
```diff
@@ -169,8 +169,8 @@ def main() -> None:
     # To make sure we can compare training and validation loss (e.g. for fine-tuning), we keep a special set
     # called "matchtrain" that uses the same data as the training set.
     val_sets = dict(
-        [("matchtrain", args.data_sets)]
-        + [(x, [x]) for x in args.val_sets]
+        # [("matchtrain", args.data_sets)]
+        [(x, [x]) for x in args.val_sets]
```
farzadab, line 173: It turns out we have some issues regarding validation and GPU stalling. Apparently we've been using the `train` set for validation by mistake.

When I fixed that I ran into some issues: the first one was fixed by changing the number of shards on HeySQuAD, but GS is still giving me issues in validation so I disabled `matchtrain` for now specially since even if I fixed GS, we might still run into issues as we change `--data-sets`, which are gonna be harder to suspect (e.g. you change train-sets and now finally your training is stuck during validation ...).

ultravox/data/datasets.py
```diff
@@ -637,7 +637,7 @@ class HeySQuADHumanDataset(QAVoiceDatasetMixin):
     def __init__(self, args: VoiceDatasetArgs) -> None:
         super().__init__(args)
         dataset = self._load_audio_dataset(
-            "yijingwu/HeySQuAD_human", split=args.split.value
+            "fixie-ai/HeySQuAD_human", split=args.split.value
```
farzadab, line 640: The only change is adjusting `num_shards` to get around the GPU stalling issue.

ultravox/training/configs/meta_config.yaml
```diff
@@ -1,4 +1,4 @@
-text_model: "meta-llama/Meta-Llama-3-8B-Instruct"
+text_model: "meta-llama/Meta-Llama-3.1-8B-Instruct"
```
farzadab, line 1: Changing default. This might be a bit pre-mature, so I don't have a strong opinion on it.

mcloud.yaml
zqhuang211, line 13: isn't it better to keep the UV_CONFIG env?

farzadab: I explained this separately. I don't know how we can control this from the outside, so it's not really doing anything for us at this point. Originally Justin added this assuming our local envs could control this but that wasn't the case IIRC.

zqhuang211: I think it is similar to UV_BRANCH, it is cleaner and easier to edit when they are separated from the rest.

farzadab: I don't like `UV_BRANCH` either, it's just a level of indirection that doesn't do anything. Though that one is not as annoying.
Also we don't use `MLFLOW_TRACKING_URI` any more.

zqhuang211: For new users, this signifies that these are typical changes you can make to run custom experiments, and it is easy to understand. The same can be said about git_repo.

ultravox/model/ultravox_model.py
```diff
@@ -16,10 +16,7 @@
 from .whisper_model_modified import WhisperEncoder as ModifiedWhisperEncoder
 
 
-class UltravoxModel(
-    transformers.LlamaPreTrainedModel,
-    transformers.GenerationMixin,
```
zqhuang211, line 21: Is transformers.GenerationMixin no longer needed?

mcloud.yaml
farzadab: Yeah I mentioned that above, but `UV_CONFIG` doesn't even do that well, since it's saying you have to use one of the configs, but we never use it like that.

If we could change it to: `UV_ARGS` and let it include all arguments passed to the Python code it would be good, but I'm not sure if that would work correctly out of the box (e.g. the whole env var could be interpreted as a single argument, not sure).

ultravox/model/ultravox_model.py
farzadab: It's apparently part of `PretrainedModel` (parent of `LlamaPretraiendModel`). Not sure if this was always the case or just a new change. That means it was being added twice for no reason.

farzadab:
Huh, the transformers version change was note pushed.

ultravox/training/train.py
zqhuang211: So this is a temporary fix for now, and we will add matchtrain back later? Maybe create an issue for the GPU stalling problem during validation?

mcloud.yaml
zqhuang211: I guess it is not a big deal. The mcloud.yaml is more about providing an example on how to train a model at this point. I will take it.

ultravox/model/ultravox_model.py
```diff
@@ -101,7 +98,7 @@ def forward(
         attention_mask: Optional[torch.Tensor] = None,
         audio_token_start_idx: Optional[torch.Tensor] = None,
         audio_token_len: Optional[torch.Tensor] = None,
-        past_key_values: Optional[Tuple] = None,
+        past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = None,
```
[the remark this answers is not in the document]
farzadab: `past_key_values` contains the key-value cache that is used for auto-regressive generation. The `.generate` method is the one that is providing us with `past_key_values`.

Previously this was a simple `Tuple`, but now it's enclosed in a Python object, the concrete implementation of which is [DynamicCache](https://huggingface.co/docs/transformers/en/internal/generation_utils#transformers.DynamicCache) and the `Tuple` is called the "legacy_cache".

I can't quite find when this change happened.

ultravox/training/configs/meta_config.yaml
zqhuang211: Maybe keep llama3 for now and consider switching to llama3.1 after running more experiments?

ultravox/training/train.py
farzadab: Done.

ultravox/training/configs/meta_config.yaml
farzadab: Okay reverting.

zqhuang211 (approved):

mcloud.yaml
farzadab: Alright, I modified this: now there's a `TRAIN_ARGS` that you can fill up instead of `UV_CONFIG`. It contains all of the args.

PR #57, opened 2024-07-25
Add GenericVoiceDataset class to support new datasets

zqhuang211:
New datasets can be supported by specifying the data_dicts parameter
    in yaml config file, e.g.,
    
    data_dicts:
        name: "clean"
        splits:
          - "train.100"
          - "train.360"
        name: "other"
        splits:
          - "train.500"
    
    See ultravox.training.config_base.py for parameter specs:
    
    class DataDictConfig(BaseModel):
        path: str  # Path to the dataset, or huggingface dataset id
        name: str = None # Name of the dataset, or huggingface dataset config/subset
        streaming: bool = True
        user_template: str = "<|audio|>"
        assistant_template: str = "{{text}}"
        transcript_template: str = "{{text}}"


ultravox/data/datasets.py
```diff
@@ -966,8 +1029,11 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
         "soda": SodaDataset,
         "dummy": LibriSpeechDummyDataset,
     }
-    name, *ext = name.split(":")
-    return DATASET_MAP[name](args, *ext)
+    if isinstance(name, config_base.DataDictConfig):
+        return GenericVoiceDataset(args, name)
```
farzadab, line 1033: I don't like the idea of having a "special" class that works completely different from the rest of the classes and has its own special input format. When you go down this path soon you're gonna have 2, 3, more special classes.

ultravox/training/config_base.py
```diff
@@ -23,6 +41,12 @@ class TrainConfig:
     do_train: bool = True
     do_eval: bool = True
 
+    # Due to simple_parsing's lack of support for containers of dataclass types,
+    # we first parse the data_dicts as a list of dictionaries. After parsing,
+    # we convert these dictionaries to DataDictConfig objects using Pydantic
+    # to enforce type constraints and validation, in the __post_init__ method.
+    data_dicts: Optional[List[Dict[str, Any]]] = None
```
farzadab, line 48: I assumed nesting was possible. I'll test this out.

ultravox/data/datasets.py
```diff
@@ -20,6 +21,7 @@
 from torch.utils import data
 
 from ultravox.data import text_proc
+from ultravox.training import config_base
```
juberti, line 24: I think we want `training` to depend on `data`, not the other way around. I'd move `DataDictConfig` here and call it `GenericDatasetConfig` or some such.

ultravox/data/datasets.py
```diff
@@ -947,6 +949,67 @@ def _get_sample(self, row) -> VoiceSample:
...
+            print(f"Error rendering template: {e}")
+            print(f"user_template: {self.user_template}")
+            print(f"assistant_template: {self.assistant_template}")
+            print(f"transcript_template: {self.transcript_template}")
+            print(f"sample keys: {list(row.keys())}")
+            raise ValueError(
+                f"Template rendering failed. Make sure all keys in the template exist in the sample."
+            ) from e
+
+        self._make_sample(
```
juberti, line 1004: Should this return the result of _make_sample? I don't understand why this has `_make_sample` and also `_get_transcribe_sample`.

juberti (commented):
To assess the generality of this approach, can you comment on which of our existing dataset classes could be replaced with this technique?

ultravox/data/datasets.py
zqhuang211: Good point. Will do.

zqhuang211:
Just by a quick look, the following existing dataset classes could be replaced using this approach. Unless we are doing something very customized, anything that can be simply handled by jinja2 template would work. 

- LibriSpeechDummyDataset
- BoolQDataset
- BoolQInputDataset
- BoolQWithExtendedAnswerDataset
- GigaSpeechDataset
- VoxPopuliDataset
- CommonVoiceDataset
- CoVoST2Dataset
- PeopleSpeechDataset

ultravox/data/datasets.py
zqhuang211: Thanks for catching this. This may explain the unstable results I have got so far.

ultravox/training/configs/llama3_whisper_s_cont_librispeech_960.yaml
```diff
@@ -0,0 +1,33 @@
...
+    user_template: "Continue the following text using less than 50 words:\n\n<|audio|>"
+    assistant_template: "{{ continuation }}"
+    transcript_template: "{{ text }}"
+  - path: "fixie-ai/librispeech_asr"
+    name: "other"
+    splits:
+      - "train.500"
+    user_template: "Continue the following text using less than 50 words:\n\n<|audio|>"
+    assistant_template: "{{ continuation }}"
+    transcript_template: "{{ text }}"
```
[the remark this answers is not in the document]
zqhuang211: I think it makes more sense to have data related settings to config. The choice of training data is just part of experiment configuration. 

two levels of configs is also good, if we are able to make a logical separation between the two.

ultravox/data/datasets.py
zqhuang211: A majority of existing dataset classes can be handled by this generic dataset class. I think it makes sense to have a predefined dataset class for the datasets that we use frequently. At the same time, this could also be achieved using a predefined config (handled by a generic dataset class) as part of the code repo.

ultravox/training/config_base.py
```diff
@@ -77,6 +101,16 @@ class TrainConfig:
...
     def __post_init__(self):
+        if self.data_dicts:
+            self.data_dicts = [
+                DataDictConfig(**data_dict) for data_dict in self.data_dicts
+            ]
+            if self.data_sets:
+                self.data_sets.extend(self.data_dicts)
+            else:
+                self.data_sets = self.data_dicts
+            del self.data_dicts
```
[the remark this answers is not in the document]
zqhuang211: This is really a hack to deal with simple_parsing's lack of support for  containers of dataclass types (not supported yet.)

ultravox/training/configs/llama3_whisper_s_cont_librispeech_960.yaml
```diff
@@ -0,0 +1,33 @@
+exp_name: "llama3_whisper_s_cont_librispeech_960"
```
farzadab, line 1: For now we have kept the configs directory very minimal.

In general my assumption is that configs in `main` should not be experimental.
IMO, PRs that want to merge configs should have different criteria that are based on comparing metrics not just like normal software-based PRs.

zqhuang211: I agree. The config here is mainly used as example of how to use the generic dataset config to load examples. Maybe configure it like the asr_tinyllama?

ultravox/training/config_base.py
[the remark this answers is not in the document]
farzadab: I understand, but I think more care should be taken as to not reduce the quality of the code.

For example, there's no need to delete anything. Maybe use `_` in the name to suggest the variable is private. Also declaring a variable is of type X when it clearly isn't in some cases seems wrong.

ultravox/data/datasets.py
farzadab: I'm not talking about functionality here (there I do think the generic class is lacking, but it can improve).

I'm talking about code structure from a purely software-engineering stand point.

IMO if you want to change the structure, you should change it once and for all, not have special cases. We don't have to preserve backward compatibility much.

farzadab:
> Just by a quick look, the following existing dataset classes could be replaced using this approach ...

Feel free to actually replace them, there's nothing sacred about the current format. The only constraint I would add is to make sure that `create_dataset` still supports on the same datasets.

That's gonna require making the `GenericVoiceDataset` more generic, but that's probably not too hard. Also you probably need to have the "dataset-specific config templates" committed somewhere that `datasets.py` can automatically load.

ultravox/data/datasets.py
zqhuang211: Agreed. Maybe we keep this feature off the main for now until we come up with something nicer.

ultravox/data/datasets.py
```diff
@@ -947,6 +949,67 @@ def _get_sample(self, row) -> VoiceSample:
...
+                )
+                for s in config.splits
+            ]
+        )
+
+        if self._args.shuffle:
+            dataset = dataset.shuffle(seed=self._args.shuffle_seed)
+
+        if config.num_samples:
+            dataset = dataset.select(range(config.num_samples))
```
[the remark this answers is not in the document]
zqhuang211: There could be situations where you want to only use certain number of samples from a dataset for training, for example to achieve some balance between different datasets.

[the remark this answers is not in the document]
farzadab: Hmm, I'm not sure how that would actually work in practice with the current code-base unless you come up with an alternative for `InterleaveDataset`.

If you set `repeat=False`, this can have an effect, but `repeat=False` in its current form is not very desirable because the composition of data changes in later batches in ways that are very hard to reason about.

All of that aside, you again need a comment for not working with MDS datasets and you also need an if condition to handle the streaming and non-streaming cases (i.e. `select` vs `take`).

That reminds me, are you sure the code works with both streaming and non-streaming datasets? IIRC `VoiceDataset` assumes that the dataset in `_init_dataset` is iterable, but I might be wrong.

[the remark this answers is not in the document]
zqhuang211: These are good points, need to look into the details.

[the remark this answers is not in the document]
farzadab: I think I'm actually wrong about `_init_dataset` expecting an `IterableDataset`. I just haven't tested it in practice. We only assume that it can be iterated on which both `Dataset` and `IterableDataset` probably support :?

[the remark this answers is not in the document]
zqhuang211: tbh I haven't given it much thought. I simply included it because the use of num_samples in TrainConfig suggests that it could be a nice feature for individual datasets too.

ultravox/data/datasets.py
farzadab: Just to be clear, I'm suggesting maybe we could change the structure for all datasets, but just keep it consistent. And the special cases get handled in one place instead of two (once when configs are loaded and once in `create_datasets`).

zqhuang211: Agreed. That would require some time to iron out all the details and corner cases.

zqhuang211:
@juberti Please ignore the changes related to experiment configs for now as I am figuring out a bug fix.

juberti:
A lot of good stuff in this PR but I think there's too much to review effectively right now. I think the right thing to do is to split it into 4 separate PRs, with this priority order:
- alt_* and KL loss
- interleave ratios
- generic dataset class 
- ds_tool changes
Happy to help do this splitting and also write tests for the Interleave class, which is getting fairly complicated here. @zqhuang211 let me know how I can best help here.

zqhuang211:
> A lot of good stuff in this PR but I think there's too much to review effectively right now. I think the right thing to do is to split it into 4 separate PRs, with this priority order:
> 
> * alt_* and KL loss
> * interleave ratios
> * generic dataset class
> * ds_tool changes
>   Happy to help do this splitting and also write tests for the Interleave class, which is getting fairly complicated here. @zqhuang211 let me know how I can best help here.

Agreed. I will break it into smaller PRs.

PR #59, opened 2024-07-26
HF Model loading fixes

farzadab:
This PR fixes the following:
1. gets rid of the unexpected weights warnings
1. makes HF model loading faster by skipping random initialization

Example of the unexpected weights warning that this PR is supposed to resolve:
```
Some weights of the model checkpoint at fixie-ai/ultravox_dev were not used when initializing UltravoxModel: ['language_model.lm_head.weight', 'language_model.model.embed_tokens.weight', 'language_model.model.layers.0.input_layernorm.weight',

...

- This IS expected if you are initializing UltravoxModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing UltravoxModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
```

### DIFF
--- a/ultravox/model/ultravox_model.py
+++ b/ultravox/model/ultravox_model.py
@@ -33,6 +33,8 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
     config_class = UltravoxConfig
     config: UltravoxConfig  # for type hinting
     _no_split_modules = ["Wav2Vec2Model", "WhisperEncoder", "LlamaDecoderLayer"]
+    _keys_to_ignore_on_load_missing = ["audio_tower.*"]
+    _keys_to_ignore_on_load_unexpected = ["audio_tower.*", "language_model.*"]
 
     def __init__(self, config: UltravoxConfig):
         super().__init__(config)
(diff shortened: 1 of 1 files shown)


ultravox/model/ultravox_model.py
```diff
@@ -33,6 +33,8 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
     config_class = UltravoxConfig
     config: UltravoxConfig  # for type hinting
     _no_split_modules = ["Wav2Vec2Model", "WhisperEncoder", "LlamaDecoderLayer"]
+    _keys_to_ignore_on_load_missing = ["audio_tower.*"]
```
farzadab, line 36: This one is technically not required, but I think having it will not hurt.

zqhuang211 (approved):

ultravox/model/ultravox_model.py
juberti: please add a comment explaining what these lines do, for posterity

juberti (approved):

ultravox/model/ultravox_model.py
farzadab: Added a three line explanation for each :)

PR #60, opened 2024-07-26
Add "without any explanation" to ST prompt

farzadab:


ultravox/data/datasets.py
```diff
@@ -848,10 +848,10 @@ class CoVoST2Dataset(VoiceDataset):
 
     # We currently don't use this dataset for training, so mainly the first prompt it ever used.
     TRANSLATE_PROMPTS = [
-        "Translate the following into {target}: <|audio|>",
-        "Translate the following into {target} language: <|audio|>",
-        "Please convert the following into {target}.\n<|audio|>",
-        "Could you translate this to {target} language?\n<|audio|>",
+        "Translate the following into {target}, without any explanation: <|audio|>",
```
juberti, line 851: Is this to just give a more diverse set of examples/

juberti (approved):

ultravox/data/datasets.py
farzadab: I wasn't sure about the prompts in 2+ positions and also a bit hypothetical since we're not even using it so I can revert.

But yeah the thought was to diversify more. At inference-time we're really strict, but more diversity is likely better, specially since with KD we won't overfit as much to the same output format.

juberti: seems reasonable, maybe add a comment to this effect.

PR #61, opened 2024-07-30
Separate files for eval logs

farzadab:
Removes eval sample logs from the main log path to declutter it and they're each written to their own specific file so that they are easier to inspect.

The new files will be stored under the `Files` tab in W&B in a separate `evals` directory.

image
image
image
image

### DIFF
--- a/ultravox/training/evaluation.py
+++ b/ultravox/training/evaluation.py
@@ -2,6 +2,7 @@ import concurrent.futures
 import dataclasses
 import functools
 import os
+import json
 from typing import List, Optional
 
 import numpy as np
@@ -170,13 +171,16 @@ def evaluate(
         print(agg_score_str)
 
         if log_dir:
+            eval_details = {
+                "score": average,
+                "confidence_interval": std,
+                "task_info": dataclasses.asdict(task),
+                "samples": [
+                    {**dataclasses.asdict(sample), "score": score}
+                    for sample, score in zip(output_samples, scores)
+                ],
+            }
             with open(os.path.join(log_dir, f"{task.name}.json"), "w") as f:
-                f.write(f"Task info: {str(task)}\n")
-                f.write(agg_score_str + "\n")
-                for sample, score in zip(output_samples, scores):
-                    f.write("-" * 20 + "\n")
-                    f.write(f"Q: {sample.question}\n")
-                    f.write(f"A: {sample.generated_answer}\n")
-                    f.write(f"X: {sample.expected_answer} [score: {score:.2f}]\n")
+                json.dump(eval_details, f, indent=1)
 
     return metrics
--- a/ultravox/training/evaluation.py
+++ b/ultravox/training/evaluation.py
@@ -1,8 +1,8 @@
 import concurrent.futures
 import dataclasses
 import functools
-import os
 import json
+import os
 from typing import List, Optional
 
 import numpy as np
--- a/ultravox/training/evaluation.py
+++ b/ultravox/training/evaluation.py
@@ -113,12 +113,16 @@ def evaluate(
     num_procs: int = 8,
     max_new_tokens: Optional[int] = None,
     temperature: Optional[float] = None,
-    verbose: bool = False,
+    log_dir: Optional[str] = None,
 ):
     metrics = {}
 
     world_size = int(os.environ.get("WORLD_SIZE", 1))
     local_rank = int(os.environ.get("LOCAL_RANK", 0))
+
+    if log_dir:
+        log_dir = os.path.join(log_dir, "evals")
+        os.makedirs(log_dir, exist_ok=True)
 
     for task in EVAL_SCENARIOS:
         ds_args = datasets.VoiceDatasetArgs(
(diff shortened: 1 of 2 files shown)


ultravox/training/train.py
```diff
@@ -131,6 +132,7 @@ def main() -> None:
             name=args.exp_name,
             dir="runs",
             tags=args.run_tags,
+            save_code=True,
```
farzadab, line 135: Also enabled the `save_code=True` option for W&B that lets a `diff.patch` of the current run to be stored in W&B for better future reproducibility (tbh this should only be useful for interactive runs where the current Git working tree is dirty).

ultravox/training/evaluation.py
```diff
@@ -156,21 +160,23 @@ def evaluate(
...
-        print(
-            f"Aggregate {task.metric} score for {task.dataset}: {average:.2f} ± {std:.2f}"
-        )
+        has_audio_str = "with" if task.include_audio else "without"
+        agg_score_str = f"Aggregate {task.metric} score for {task.dataset} ({has_audio_str} audio): {average:.2f} ± {std:.2f}"
+        print(agg_score_str)
+
+        if log_dir:
+            with open(os.path.join(log_dir, f"{task.name}.json"), "w") as f:
+                f.write(f"Task info: {str(task)}\n")
```
farzadab, line 174: Full task info is superior as it gives the full picture of the current evaluation scenario.

ultravox/training/train.py
```diff
@@ -302,7 +304,7 @@ def main() -> None:
             num_procs=args.eval_num_procs,
             num_samples=args.eval_num_samples,
             max_new_tokens=args.eval_max_new_tokens,
-            verbose=True,
+            log_dir=wandb.run.dir if wandb.run is not None else str(args.logs_dir),
```
farzadab, line 307: I wanted to keep to `logs_dir`, but I couldn't find a way to force W&B to use this path. I didn't spend extra time on it either since we only use W&B at the moment.

ultravox/training/evaluation.py
```diff
@@ -156,21 +160,23 @@ def evaluate(
...
+                f.write(agg_score_str + "\n")
+                for sample, score in zip(output_samples, scores):
+                    f.write("-" * 20 + "\n")
+                    f.write(f"Q: {sample.question}\n")
+                    f.write(f"A: {sample.generated_answer}\n")
+                    f.write(f"X: {sample.expected_answer} [score: {score:.2f}]\n")
```
zqhuang211, line 180: For logging the results, how does this compare to simply printing in jsonl format with the following keys: Q, A, X, S or question, answer, expected, score? Using jsonl format makes it easier if we need to parse the result.

ultravox/training/train.py
zqhuang211: maybe simplify it to:

`log_dir=wandb.run.dir if wandb.run else str(args.logs_dir)`

ultravox/training/evaluation.py
farzadab: The focus was on keeping human readability and matching existing formats.

I thought about JSONL, but having all of this info in one line is gonna make it unreadable.
With JSON we might be able to get something, but still we'd have to compromise readability a bit.

The question is: how important is it to read these programatically vs by humans.
We already have a snippet that can parse this, but if the programmatic readability becomes more important and we want to do it easily multiple times then we can do JSON.

zqhuang211: We can easily use tools like jq to pretty-print jsonl files on demand.

farzadab: Wait, by "on demand" do you mean you want to do this in the command-line?
The point of logging is to make it as easy as possible to use. Downloading, then using `jq` seems like two steps too many for just seeing the output of evals.

I don't know if W&B gives you a CLI for interacting with Files. I see a Python API, but that's not gonna be convenient without extra tooling support.

zqhuang211: As long as we can manually download the files, it is okay. We will not be checking eval outputs for every experiment run. Our focus will mainly be on the aggregated scores that are already there in w&b. The eval output logging will primarily be used for analysis.

farzadab: I disagree. Outputs should be easily inspectable.

Remember [this Slack thread](https://fixie-ai.slack.com/archives/C06CXPY070V/p1722032451121769?thread_ts=1722032121.134989&cid=C06CXPY070V) and the exploration to compare sentence-bleu vs corpus-bleu?

That whole exploration need not have occurred if the actual outputs were glanced at first.

farzadab: Anyway, I'll do JSON with `indent=1`.

juberti: I agree that human readability is worth preserving, JSONLs are notoriously hard to eyeball.

zqhuang211: That's what I meant by analysis. We downloaded the output files, parsed the results, computed and verified scores, and then realized that the problem was with prompting. 

We want make it easier to perform analysis, either programmatically or by human, and it goes beyond functionalities provided on wandb.

zqhuang211: If we wan to have a balance, maybe choose json instead of jsonl.

farzadab: > That's what I meant by analysis. We downloaded the output files, parsed the results, computed and verified scores, and then realized that the problem was with prompting.

Again, going back to the example (not trying to make this personal, just using this as example): you didn't do it on the first time, which proves my point: people do what's easiest. If you knew you could just look at logs and they were easy to check, you would've done that first.

farzadab: It's JSON now.

zqhuang211: You are right. Actually I didn’t catch the problem when I first looked at the downloaded files. I reran bleu calculation, verified your results, and then started to look for other causes. Anything we do to make it easier to spot errors would be helpful.

ultravox/training/evaluation.py
```diff
@@ -113,13 +114,17 @@ def evaluate(
     num_procs: int = 8,
     max_new_tokens: Optional[int] = None,
     temperature: Optional[float] = None,
-    verbose: bool = False,
+    log_dir: Optional[str] = None,
```
zqhuang211, line 117: should this be enabled by default?

farzadab: I don't understand the question. Are you suggesting it should be and it's not. Or are you suggesting it shouldn't be enabled and it is?

zqhuang211: I assume that when evaluate() is called, by default the caller will pass in a log_dir, correct? If that's the case, please ignore the comment. I want to make sure that by default logging of eval results are enabled.

zqhuang211: I saw it was set properly in trainer.py. This is all good.

zqhuang211 (approved):

PR #62, opened 2024-07-31
[ds_tool] Tools with Audio

farzadab:
This PR allow `ds_tool` to TTS multiple audios with the same voice.
It also adds configs used to create [`fixie-ai/tools-audio`](https://huggingface.co/datasets/fixie-ai/tools-audio) (note: this dataset is likely to be merged into the original `fixie-ai/tools`).

### DIFF
--- a/ultravox/tools/ds_tool/caching.py
+++ b/ultravox/tools/ds_tool/caching.py
@@ -1,7 +1,6 @@
 import hashlib
 import json
 import os
-import random
 from typing import List, Optional, Union, overload
 
 import openai
--- a/ultravox/tools/ds_tool/caching.py
+++ b/ultravox/tools/ds_tool/caching.py
@@ -53,15 +53,12 @@ class CachingTtsWrapper:
         self, text: Union[str, List[str]], voice: Optional[str] = None
     ) -> Union[bytes, List[bytes]]:
         text_hash = hashlib.sha256(str(text).encode()).hexdigest()
+        voice = self._client.resolve_voice(voice)
 
         if isinstance(text, list):
-            if voice == tts.RANDOM_VOICE_KEY and hasattr(self._client, "ALL_VOICES"):
-                voice = random.Random(int(text_hash, 16)).choice(
-                    self._client.ALL_VOICES
-                )
             return [self.tts(t, voice) for t in text]
 
-        path = os.path.join(self._base_path, voice or "default")
+        path = os.path.join(self._base_path, voice)
         os.makedirs(path, exist_ok=True)
 
         cache_path = os.path.join(path, f"{text_hash}.wav")
--- a/ultravox/tools/ds_tool/caching.py
+++ b/ultravox/tools/ds_tool/caching.py
@@ -1,7 +1,8 @@
 import hashlib
 import json
 import os
-from typing import Optional
+import random
+from typing import List, Optional, Union, overload
 
 import openai
 
@@ -42,9 +43,25 @@ class CachingTtsWrapper:
         self._client = client
         self._base_path = os.path.join(".cache/ds_tool/tts", provider)
 
-    def tts(self, text: str, voice: Optional[str] = None) -> bytes:
+    @overload
+    def tts(self, text: str, voice: Optional[str] = None) -> bytes: ...
+
+    @overload
+    def tts(self, text: List[str], voice: Optional[str] = None) -> List[bytes]: ...
+
+    def tts(
+        self, text: Union[str, List[str]], voice: Optional[str] = None
+    ) -> Union[bytes, List[bytes]]:
+        text_hash = hashlib.sha256(str(text).encode()).hexdigest()
+
+        if isinstance(text, list):
+            if voice == tts.RANDOM_VOICE_KEY and hasattr(self._client, "ALL_VOICES"):
+                voice = random.Random(int(text_hash, 16)).choice(
+                    self._client.ALL_VOICES
+                )
+            return [self.tts(t, voice) for t in text]
+
         path = os.path.join(self._base_path, voice or "default")
-        text_hash = hashlib.sha256(text.encode()).hexdigest()
         os.makedirs(path, exist_ok=True)
 
         cache_path = os.path.join(path, f"{text_hash}.wav")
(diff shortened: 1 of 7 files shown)


ultravox/tools/ds_tool/caching.py
```diff
@@ -42,9 +43,25 @@ def __init__(self, client: tts.Client, provider: str):
...
-    def tts(self, text: str, voice: Optional[str] = None) -> bytes:
+    @overload
+    def tts(self, text: str, voice: Optional[str] = None) -> bytes: ...
+
+    @overload
+    def tts(self, text: List[str], voice: Optional[str] = None) -> List[bytes]: ...
+
+    def tts(
+        self, text: Union[str, List[str]], voice: Optional[str] = None
+    ) -> Union[bytes, List[bytes]]:
```
farzadab, line 54: TTS cache diverged from the original TTS interface. I could've modified the original, but it didn't seem necessary. Also doing it at a layer above/below seemed too complicated since the voices needed to be the same.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -20,7 +21,8 @@
 class TtsTask:
     implementation: str = simple_parsing.field(default="azure", alias="-i")
     # Column name containing the text to convert to audio. It can be a Jinja variable expression.
-    column_name: str = simple_parsing.field(default="question", alias="-c")
+    template: str = simple_parsing.field(alias="-T")
+    json_mode: bool = simple_parsing.field(default=False, alias="-j")
```
farzadab, line 25: A `column_name` wasn't really enough. Now these two variables are shared between the two types and can potentially be moved to the parent class. I didn't do it yet, since future tasks might not follow the same structure.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -112,7 +127,7 @@ def _map_sample(self, sample):
             ) from e
 
         if self.json_mode:
-            turns = json.loads(rendered)
+            turns = yaml.safe_load(rendered)
```
farzadab, line 130: YAML is a superset of JSON. The issue was that I couldn't figure out how to remove the final comma (`,`) which JSON really strict on. Moving to YAML seemed like and given that we verify the type assumptions, I didn't see why it could hurt.

It does make the naming a bit awkward though: `json_mode` uses YAML. An option is to also change the name to `yaml_mode`.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -130,9 +145,9 @@ def _map_sample(self, sample):
 
 
 # This script is used to either generate audio samples from text using a TTS model, or to generate text samples using a text generation model.
-#   just ds_tool tts -d google/boolq -u fixie-ai/boolq-audio -c question -a audio --token $HF_WRITE_TOKEN
-#   just ds_tool textgen -d fixie-ai/boolq-audio -u fixie-ai/bar -c explanation -b https://api.fireworks.ai/inference/v1 -k $FIREWORKS_API_KEY -m accounts/fireworks/models/llama-v3-8b-instruct
-#   just ds_tool textgen -d ylacombe/expresso -u fixie-ai/expresso -c continuation -T @expresso_template.txt
+#   just ds_tool tts -d google/boolq -u fixie-ai/boolq-audio -T {{question}} -a audio --token $HF_WRITE_TOKEN
```
farzadab, line 148: I can support the plain column_name format (e.g. `question` instead of `{{question}}`), but that seemed more hacky than necessary.

ultravox/tools/ds_tool/user_messages.jinja
```diff
@@ -0,0 +1,7 @@
+[
```
juberti, line 1: Should you still have the square braces if you're using YAML?

ultravox/tools/ds_tool/ds_tool.py
juberti: Agree this is a bit awkward, but not sure what the alternative is, I agree the trailing comma issue is annoying. (I guess you could have some sort of second parameter to `json_dump` to control whether to add a comma or not, but even that is ugly.) 

I think it's fine to keep `json_mode` as the name for now though as plain JSON is still supported.

ultravox/tools/ds_tool/ds_tool.py
juberti: I tend to think we'll want to hoist these, but we can certainly do it later.

juberti (approved):
LG with nits

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -35,32 +37,45 @@ def __post_init__(self):
             provider=self.implementation,
```
zqhuang211, line 37: why use `provider` in the TTS wrapper but `implementation` in the actually TTS client?

ultravox/tools/ds_tool/caching.py
farzadab: To verify, do you mean not to "overload" the `tts` function or just let the underlying class handle resolving the voice name?
Asking since the comment is about the latter but the lines highlighted are about the former.

farzadab: Added `resolve_voice`.

ultravox/tools/ds_tool/ds_tool.py
farzadab: No real reason. Renamed `provider` to `implementation` to make them move consistent.

ultravox/tools/ds_tool/user_messages.jinja
farzadab: Hmm, I think it's possible to do it without the braces, but then I'll have to be really careful about number of tabs/newlines which is not easy to figure out in Jinja.

With the braces it's way more forgiving on those types of things.

ultravox/tools/ds_tool/caching.py
juberti: The main thing I was concerned about was the details of the random voice selection leaking out of the tts impl code, which you've addressed.

PR #63, opened 2024-08-02
Support KL loss

zqhuang211:
This PR implements the (response) KL loss described in the BLSP-KD paper: https://arxiv.org/abs/2405.19041

### DIFF
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,6 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  # temperary for training verfication
   UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
-  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper.yaml
+  UV_BRANCH: zhuang/add-kl-loss
+  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: zhuang/add-kl-loss
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -3,7 +3,7 @@ name: ultravox
 image: mosaicml/composer:latest
 compute:
   gpus: 8
-  cluster: r7z22p1
+  cluster: r14z3p1
 integrations:
   - integration_type: git_repo
     git_repo: fixie-ai/ultravox
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -14,5 +14,5 @@ command: >-
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   # temperary for training verfication
-  UV_BRANCH: zhuang/add-kl-loss
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
+  # temperary for training verfication
+  UV_BRANCH: zhuang/add-kl-loss
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -69,9 +69,27 @@ logging.getLogger("streaming.base.dataset").setLevel(logging.ERROR)
 
 @dataclasses.dataclass
 class DataCollatorForSeq2SeqWithAudio(transformers.DataCollatorForSeq2Seq):
+    # when enabled, the alt_input_ids, alt_attention_mask, and alt_labels fields are used for computing the KL loss in UltravoxModel
+    include_alt_input: bool = False
+
     def __call__(self, features, *args, **kwargs):
         audio_values = [f.pop("audio_values", None) for f in features]
+        if self.include_alt_input:
+            # these fields are hard-coded in the transformer data collator, so they need special handling before calling the super method
+            alt_features = [
+                {
+                    "input_ids": f.pop("alt_input_ids"),
+                    "attention_mask": f.pop("alt_attention_mask"),
+                    "labels": f.pop("alt_labels"),
+                }
+                for f in features
+            ]
         batch = super().__call__(features, *args, **kwargs)
+        if self.include_alt_input:
+            alt_batch = super().__call__(alt_features, *args, **kwargs)
+            batch["alt_input_ids"] = alt_batch["input_ids"]
+            batch["alt_attention_mask"] = alt_batch["attention_mask"]
+            batch["alt_labels"] = alt_batch["labels"]
 
         # Pad the last dimension of all audio_values to the same length, with 0s on the right.
         if audio_values and audio_values[0] is not None:
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -70,11 +70,11 @@ logging.getLogger("streaming.base.dataset").setLevel(logging.ERROR)
 @dataclasses.dataclass
 class DataCollatorForSeq2SeqWithAudio(transformers.DataCollatorForSeq2Seq):
     # when enabled, the alt_input_ids, alt_attention_mask, and alt_labels fields are used for computing the KL loss in UltravoxModel
-    include_alt_input: bool = False
+    include_alt_fields: bool = False
 
     def __call__(self, features, *args, **kwargs):
         audio_values = [f.pop("audio_values", None) for f in features]
-        if self.include_alt_input:
+        if self.include_alt_fields:
             # these fields are hard-coded in the transformer data collator, so they need special handling before calling the super method
             alt_features = [
                 {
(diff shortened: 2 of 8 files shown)


zqhuang211:
I am running model training to verify training works properly.

ultravox/model/data_processing.py
```diff
@@ -91,8 +110,33 @@ def _process(self, sample: datasets.VoiceSample) -> Dict[str, Any]:
             )["input_ids"].shape[-1]
             labels[:input_text_len] = -100
 
+            if self.include_alt_input:
+                alt_input_text = input_text.replace(
```
juberti, line 114: Can this be simplified? The assistant message should have the same length for both input types, so can we simply do `alt_input_text_len = len(alt_text) - (len(text) - len(input_text))`?

zqhuang211:
Model training finished as expected.

ultravox/model/ultravox_model.py
```diff
@@ -75,6 +78,10 @@ def get_decoder(self):
     def tie_weights(self):
         return self.language_model.tie_weights()
 
+    def set_loss_config(self, loss_config: Optional[LossConfig] = None):
```
juberti, line 81: Why `Optional`?

ultravox/model/ultravox_model.py
```diff
@@ -162,8 +173,40 @@ def forward(
             past_key_values=past_key_values,
             **kwargs,
         )
-
-        return lm_output
+        if self.training:
+            if self.loss_config.loss_function == LossFunction.CrossEntropy:
+                return lm_output
+            elif self.loss_config.loss_function == LossFunction.KL_Divergence:
+                # compute the teacher (text-only) model's distribution
```
juberti, line 180: suggest hoisting this into a `self._compute_kl_loss()` function to try to keep the core `forward()` logic clear

ultravox/model/ultravox_config.py
```diff
@@ -20,6 +21,20 @@ class LoraConfigSimplified:
...
+class LossFunction(str, Enum):
+    CrossEntropy = "ce"
+    KL_Divergence = "kl"
+
+
+@dataclasses.dataclass
+class LossConfig:
+    loss_function: LossFunction = LossFunction.KL_Divergence
+    kl_temperature: float = 2.0
+
```
juberti, line 33: suggest making this a `@property`

ultravox/training/train.py
```diff
@@ -155,6 +159,8 @@ def main() -> None:
                 )
 
     model.print_trainable_parameters()
+    # loss_config needs to be passed separately just for model training
+    model.set_loss_config(args.loss_config)
```
juberti, line 163: suggest moving this before line 128 so all model init happens in one place. Also, suggest only calling `set_loss_config` if `args.loss_config` isn't None, to avoid the need for `Optional`

juberti (commented):
Thanks for splitting this out, mostly minor comments.

mcloud.yaml
```diff
@@ -13,5 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
-  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper.yaml
+  UV_BRANCH: zhuang/add-kl-loss
```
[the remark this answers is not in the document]
zqhuang211: Reverted after training run passed.

ultravox/data/datasets.py
```diff
@@ -69,9 +69,27 @@
 
 @dataclasses.dataclass
 class DataCollatorForSeq2SeqWithAudio(transformers.DataCollatorForSeq2Seq):
+    # when enabled, the alt_input_ids, alt_attention_mask, and alt_labels fields are used for computing the KL loss in UltravoxModel
+    include_alt_input: bool = False
```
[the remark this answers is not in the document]
zqhuang211: We need to prepare alternative versions of the fields `input_ids`, `attention_mask`, and `labels` in a batch for text input. I struggled with finding the right name for the flag. I wanted to emphasize `alt_` to match the `alt_*` fields. However, `input_alt_input` does not capture the `alt_attention_mask` and `alt_labels` fields. Similarly, `include_text_input` does not account for the `alt_` prefix or the other fields.

Perhaps we could use `include_alt_fields`?

ultravox/model/data_processing.py
```diff
@@ -45,6 +49,9 @@ def _process(self, sample: datasets.VoiceSample) -> Dict[str, Any]:
             sample.messages, tokenize=False
         )
 
+        if self.include_alt_input:
+            alt_text = text.replace("<|audio|>", sample.audio_transcript or "")
```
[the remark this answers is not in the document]
zqhuang211: The `_get_sample()` method in `datasets.py` would be a more appropriate place to handle this logic. I think it’s acceptable to support empty audio transcripts in this context, as an audio file might simply contain silence or noise.

ultravox/model/data_processing.py
```diff
@@ -65,9 +72,21 @@ def _process(self, sample: datasets.VoiceSample) -> Dict[str, Any]:
...
+                audio=None,
+                return_tensors="pt",
+            )
+            alt_input_ids = alt_inputs["input_ids"].squeeze_(0)
+            alt_inputs["attention_mask"].squeeze_(0)
+
         # No need to shift the labels as the model does it internally
         labels = input_ids.clone()
 
+        if self.include_alt_input:
```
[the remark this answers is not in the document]
zqhuang211: I debated this. Keep them separate is easier for implementing the parallels, but I agree it is probably better for readability. To be updated.

ultravox/model/data_processing.py
zqhuang211: I think you are right, though the computation should be based on tokens, not text.

`alt_input_text_len = input_text_len + len(alt_input_ids) - len(input_ids)`

ultravox/model/ultravox_model.py
zqhuang211: This is because `loss_config` defaults to None in `config_base.py`, but it shouldn't be None during training. I will update `config_base.py` to make `loss_confg` to have a non-None default value, then we can get rid of Optional here.

ultravox/data/datasets.py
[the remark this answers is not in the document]
juberti: Isn't the `alt` prefix something that we added? That is, could we just make a global change to use `text` or some similar prefix? 

Or, if we want to keep `alt`, perhaps `include_alt_modality` suggesting that we're trying to send multiple modalities to the model for the same input.

ultravox/model/data_processing.py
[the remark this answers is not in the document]
juberti: My concern was more that if there was some logic error that led to a `None` transcript, we should probably handle that rather than silently inserting an empty transcript. Even if the audio is silent, there should still be a valid "" transcript.

ultravox/model/data_processing.py
juberti: ah yeah, good catch. maybe `alt_input_token_len` then to avoid future confusion.

ultravox/model/ultravox_model.py
zqhuang211: Makes sense. Will update

ultravox/model/data_processing.py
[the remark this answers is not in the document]
zqhuang211: Agreed.

ultravox/model/data_processing.py
zqhuang211: right, i will update `input_text_len` to `input_token_len` as well.

ultravox/model/ultravox_config.py
zqhuang211: Sounds like the right thing to do. Please suggest how to do it.

ultravox/training/train.py
zqhuang211: Done. args.loss_config is updated to always have the LossConfig type.

ultravox/model/ultravox_config.py
juberti: This should be sufficient:

```suggestion
@property
```

ultravox/model/data_processing.py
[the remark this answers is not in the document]
zqhuang211: I had to keep the `or ""` statement for now, otherwise some samples cause training failure. Will revisit this later as we start to clean up `datasets.py`

ultravox/model/ultravox_model.py
zqhuang211: It turns out that I have to use Optional[LossConfig] in config_base.py, otherwise it will not pass `just check`.  I updated `train.py` to only invoke `set_loss_config` when `args.loss_config is not None`

zqhuang211:
@juberti please review the latest version when you have a chance.

juberti (approved):

ultravox/model/ultravox_model.py
```diff
@@ -162,8 +206,25 @@ def forward(
             past_key_values=past_key_values,
             **kwargs,
         )
-
-        return lm_output
+        if self.training:
```
farzadab, line 209: Hmm, why gate this with `self.training`?
Do you explicitly want the loss to be different in training vs validation?

Is is because you want the validation loss to be comparable to other runs?

ultravox/training/configs/llama3_whisper_kd.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# Temporarily remove heysquad_human from val_sets as it causes the training to fail.
+val_sets: ["anyinstruct", "soda", "peoplespeech"]
+
+# Load continuation training data using GenericVoiceDataset.
+# Select 100,000 samples from both clean and other subsets.
+# Mix clean and other subsets in a 2:1 ratio for batch creation.
+# Total training steps: 4000, with each step 4x8=32 samples.
+batch_size: 4
+max_steps: 1000
```
farzadab, line 21: Doesn't seem like the comments are reflecting the config itself? E.g. `max_steps` is 1000 not 4000. There aren't 100K samples being used here, etc. So I suggest removing them.

mcloud.yaml
```diff
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
+  # temperary for training verfication
   UV_BRANCH: main
-  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper.yaml
+  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
```
farzadab, line 18: Nit: If this is temporary, then make sure to remove it from PR :?

farzadab (approved):

ultravox/data/datasets.py
```diff
@@ -433,15 +451,13 @@ class AnyInstructDataset(VoiceDataset):
...
-        # The last 7 samples are missing audio files, so we exclude them.
-        NUM_SAMPLES = 108193 - 7
         super().__init__(args)
         dataset = datasets.load_dataset(
             "json",
             "anyinstruct",
             data_files="https://huggingface.co/datasets/fnlp/AnyInstruct/resolve/main/speech_conv/metadata.jsonl",
             split="train",
-        ).select(range(NUM_SAMPLES))
+        )
```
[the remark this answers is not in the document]
juberti: Apparently the dataset has been updated? Although we should verify that the speech data is still in sync with our local copy.

ultravox/model/data_processing.py
```diff
@@ -84,15 +88,40 @@ def _process(self, sample: datasets.VoiceSample) -> Dict[str, Any]:
...
+            alt_labels = alt_input_ids.clone()
+            if not self.train_on_inputs:
+                alt_input_token_len = (
+                    input_token_len + len(alt_input_ids) - len(input_ids)
+                )
+                alt_labels[:alt_input_token_len] = -100
+
+            inputs["alt_input_ids"] = alt_input_ids
+            inputs["alt_attention_mask"] = alt_inputs["attention_mask"]
+            inputs["alt_labels"] = alt_labels
```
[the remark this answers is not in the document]
zqhuang211: We will look into this later.

ultravox/data/datasets.py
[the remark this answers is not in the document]
zqhuang211: The dataset on huggingface was updated

PR #65, opened 2024-08-03
add generic dataset

zqhuang211:
Add GenericVoiceDataset to support new datasets via yaml config, e.g., 

```
data_dicts:
    name: "clean"
    splits:
      - "train.100"
      - "train.360"
    num_samples: 100_000
    name: "other"
    splits:
      - "train.500"
    num_samples: 100_000
```

See ultravox.training.configs.llama3_whisper_kd.yaml for example experiment configuration.

### DIFF
--- a/ultravox/data/datasets.py
+++ b/ultravox/data/datasets.py
@@ -9,6 +9,7 @@ import tempfile
 from typing import Any, Callable, Dict, List, Optional, Sequence
 
 import datasets
+import jinja2
 import librosa
 import numpy as np
 import requests
@@ -20,6 +21,7 @@ import transformers
 from torch.utils import data
 
 from ultravox.data import text_proc
+from ultravox.training import config_base
 
 SAMPLE_RATE = 16000
 
@@ -727,6 +729,7 @@ class SlueSQA5Dataset(QAVoiceDatasetMixin):
         )
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
 class LibriSpeechDataset(VoiceDataset):
     """
     LibriSpeech is a corpus of approximately 1000 hours of 16kHz read
@@ -761,6 +764,7 @@ class LibriSpeechDataset(VoiceDataset):
         return self._get_transcribe_sample(row, tproc=text_proc.format_asr_text)
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
 class GigaSpeechDataset(VoiceDataset):
     """
     GigaSpeech is an evolving, multi-domain English speech recognition corpus
@@ -780,6 +784,7 @@ class GigaSpeechDataset(VoiceDataset):
         return self._get_transcribe_sample(row, tproc=text_proc.format_asr_text)
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
 class VoxPopuliDataset(VoiceDataset):
     """
     VoxPopuli is a large-scale multilingual speech corpus for representation learning,
@@ -799,6 +804,7 @@ class VoxPopuliDataset(VoiceDataset):
         return self._get_transcribe_sample(row, tcol="raw_text")
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
 class CommonVoiceDataset(VoiceDataset):
     """
     The Common Voice dataset consists of a unique MP3 and corresponding text file
@@ -821,6 +827,7 @@ class CommonVoiceDataset(VoiceDataset):
         return self._get_transcribe_sample(row, tcol="sentence")
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
 class CoVoST2Dataset(VoiceDataset):
     """
     CoVoST 2 is a large-scale multilingual speech translation corpus covering translations from 21 languages into English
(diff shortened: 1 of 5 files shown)


ultravox/training/config_base.py
```diff
@@ -3,14 +3,32 @@
...
 
 from ultravox.model import ultravox_config
 
 
+class DataDictConfig(BaseModel):
+    # Path to the dataset, or huggingface dataset id
+    path: str
+    # Name of the dataset, or huggingface dataset config/subset
+    name: Optional[str] = None
+    splits: List[str] = dataclasses.field(default_factory=list)
```
juberti, line 20: Do we need to set up different splits for train and validation?

ultravox/training/config_base.py
```diff
@@ -20,6 +38,15 @@ class TrainConfig:
     # audio encoder model to use
     audio_model: str
 
+    # The data_dicts field complements data_sets, allowing for the inclusion of
+    # new datasets in the config.
+    #
+    # Due to simple_parsing's lack of support for containers of dataclass types,
```
juberti, line 44: Is this true? I thought we had managed to solve this previously: https://github.com/lebrice/SimpleParsing/blob/master/examples/container_types/README.md

ultravox/training/config_base.py
```diff
@@ -3,14 +3,32 @@
...
+from typing import Any, Dict, List, Optional
 
 import simple_parsing
 import torch
+from pydantic import BaseModel
 
 from ultravox.model import ultravox_config
 
 
+class DataDictConfig(BaseModel):
```
juberti, line 15: I think we might want this to live in ultravox/data so that we can integrate it into our other tooling, and also write tests for it.

ultravox/training/config_base.py
```diff
@@ -80,6 +107,16 @@ class TrainConfig:
...
 
     def __post_init__(self):
+        if self.data_dicts:
+            self.data_dicts = [
+                DataDictConfig(**data_dict) for data_dict in self.data_dicts
+            ]
+            if self.data_sets:
+                self.data_sets.extend(self.data_dicts)
+            else:
+                self.data_sets = self.data_dicts
```
farzadab, line 117: Why do you need the `if` condition? `self.data_sets` can never be None.

ultravox/training/config_base.py
```diff
@@ -80,6 +107,16 @@ class TrainConfig:
...
+            del self.data_dicts
```
farzadab, line 118: Why remove `self.data_dicts`? It just breaks the interface (not that it didn't before).

ultravox/training/config_base.py
```diff
@@ -80,6 +107,16 @@ class TrainConfig:
     loss_config: Optional[ultravox_config.LossConfig] = None
 
     def __post_init__(self):
+        if self.data_dicts:
+            self.data_dicts = [
+                DataDictConfig(**data_dict) for data_dict in self.data_dicts
+            ]
+            if self.data_sets:
+                self.data_sets.extend(self.data_dicts)
```
farzadab, line 115: nit and I guess we're not fixing it in this PR: I'm still not happy with this format. For example, you have to know that `data_dicts` get added on the right which is arbitrary and all the type hints are wrong.

ultravox/data/datasets.py
```diff
@@ -821,6 +827,7 @@ def _get_sample(self, row) -> VoiceSample:
         return self._get_transcribe_sample(row, tcol="sentence")
 
 
+# TODO: this dataset can be replaced with GenericVoiceDataset and will be removed/updated in the future.
```
farzadab, line 830: Nit: I don't think these TODO comments are entirely correct. For example the random `PROMPTS` and `CODE_TO_LANG` are not replaceable with the current version.

ultravox/training/config_base.py
zqhuang211: If we don't use the generic dataset class, the training and validation datasets are specified in the config yaml file as follows:

```
data_sets: ["gigaspeech"]
val_sets: ["heysquad_human", "anyinstruct", "soda", "peoplespeech"]
```

With the generic dataset class, the additional training datasets can be specified by:

```
data_dicts:
    name: "clean"
    splits:
      - "train.100"
      - "train.360"
    num_samples: 100_000
    name: "other"
    splits:
      - "train.500"
    num_samples: 100_000
```

Similar to the use of `--data_sets`, the datasets (and their corresponding splits) specified in `--data_dicts` are used for training. We could also include a `--val_dicts` option to specify additional validation datasets in a similar manner, but this feature is not yet implemented.

I believe it would be better to use a YAML configuration to specify all the details for training, validation, and evaluation datasets (with defaults specified in the default config YAML). We can address this in a future PR, as it requires extensive testing to support all existing datasets.

ultravox/training/config_base.py
zqhuang211: Yes, the issue is specific to containers of dataclass types. The error message says something like "support for containers of dataclass types is not supported yet"

ultravox/training/config_base.py
zqhuang211: I think you are right.

ultravox/training/config_base.py
zqhuang211: It is a hack, will update it with a comment.

ultravox/data/datasets.py
zqhuang211: These random prompts will still be used, but the corresponding datasets could be specified using the logic following the (updated) GenericVoiceDataset class, so that we have a unified way of specifying datasets for training, validation, and evaluation.

ultravox/training/config_base.py
zqhuang211: We will address this in the right way in a later PR and you will be happy :-)

PR #69, opened 2024-08-05
Fix TypeError: non-default argument 'template' follows default argument, and filter out audio

liPatrick:
1. Fixes TypeError: non-default argument 'template' follows default argument
2. Filters out audio before passed into jinja template to prevent memory spikes when using ds_tool.py

### DIFF
--- a/ultravox/tools/ds_tool/continuation.jinja
+++ b/ultravox/tools/ds_tool/continuation.jinja
@@ -1,3 +1,3 @@
 Continue the following text using less than 50 words:
 
-{{ text_proc.format_asr_text(sentence) }}
+{{ text_proc.format_asr_text(text) }}
--- a/ultravox/tools/ds_tool/continuation.jinja
+++ b/ultravox/tools/ds_tool/continuation.jinja
@@ -1,3 +1,3 @@
 Continue the following text using less than 50 words:
 
-{{ text_proc.format_asr_text(text) }}
+{{ text_proc.format_asr_text(sentence) }}
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -113,7 +113,9 @@ class TextGenerationTask:
     def _map_sample(self, sample):
         # e.g., {{ text }} or {{ text_proc.format_asr_text(text) }}
         try:
-            filtered_sample = {"sentence": sample["sentence"]}
+            # We need to filter out the audio before the sample is passed into the jinja template
+            # or it will get loaded into memory and spike usage.
+            filtered_sample = {k: v for k, v in sample.items() if k != "audio"}
             rendered = jinja2.Template(
                 self.template, undefined=jinja2.StrictUndefined
             ).render(**filtered_sample, json_dump=json.dumps, text_proc=text_proc)
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -19,9 +19,8 @@ chat_client: caching.CachingChatWrapper
 
 @dataclasses.dataclass
 class TtsTask:
-    template: str = simple_parsing.field(alias="-T")
     implementation: str = simple_parsing.field(default="azure", alias="-i")
-    # Column name containing the text to convert to audio. It can be a Jinja variable expression.
+    template: str = simple_parsing.field(default=None, alias="-T")
     json_mode: bool = simple_parsing.field(default=False, alias="-j")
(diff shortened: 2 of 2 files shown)


ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -19,9 +19,9 @@
 
 @dataclasses.dataclass
 class TtsTask:
+    template: str = simple_parsing.field(alias="-T")
     implementation: str = simple_parsing.field(default="azure", alias="-i")
     # Column name containing the text to convert to audio. It can be a Jinja variable expression.
```
zqhuang211, line 24: Move this comment line up before the implementation variable?

juberti: Actually this comment refers to code that no longer exists, it can be removed.
```suggestion
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -19,9 +19,9 @@
 
 @dataclasses.dataclass
 class TtsTask:
+    template: str = simple_parsing.field(alias="-T")
```
juberti, line 22: Let's leave this where it was and just add `default=None` like the other similar fields

zqhuang211: Does it make sense for template to be `None`?

ultravox/tools/ds_tool/continuation.jinja
```diff
@@ -1,3 +1,3 @@
 Continue the following text using less than 50 words:
 
-{{ text_proc.format_asr_text(text) }}
+{{ text_proc.format_asr_text(sentence) }}
```
farzadab, line 3: nit: this is dataset dependent, so there's not much point changing it here.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -112,12 +111,12 @@ def map_split(
         )
 
     def _map_sample(self, sample):
-        # using a Jinja template for some added flexibility, template can include variables and functions
         # e.g., {{ text }} or {{ text_proc.format_asr_text(text) }}
         try:
+            filtered_sample = {"sentence": sample["sentence"]}
```
farzadab, line 116: Hmm, this won't work. There are many other keys in the sample that some of the other templates make use of.

Maybe let's do an "exclude" list, with a regex like `.*audio.*` instead? We can also take the regex as an argument from the user instead of having it be fixed.

farzadab: Also it's best if you add a comment so that explains why the sample was filtered (otherwise someone might change it back later on).

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -112,12 +111,12 @@ def map_split(
         )
 
     def _map_sample(self, sample):
-        # using a Jinja template for some added flexibility, template can include variables and functions
```
farzadab, line 115: nit: comment removed.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -20,8 +20,7 @@
 @dataclasses.dataclass
 class TtsTask:
     implementation: str = simple_parsing.field(default="azure", alias="-i")
-    # Column name containing the text to convert to audio. It can be a Jinja variable expression.
```
farzadab, line 23: nit: why remove the comment?

farzadab:
Thanks for the fixes btw Pat.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: https://github.com/fixie-ai/ultravox/pull/69#discussion_r1704781138


Apprently it refers to code that no longer exists

farzadab: You're right, it should say something like: "Jinja template for the text that needs to be converted to audio".

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Think I might do an excludes fields list, because afraid we might accidently exclude a field we want with regex matching. Ie, a column might be called audio_transcript.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: default=None doesnt pass the checks. I just moved it in the front.

juberti: ah yeah, this is non-optional. LG.

juberti (approved):

PR #70, opened 2024-08-06
Add weighted sampling in InterleaveDataset

zqhuang211:
Update InterleaveDataset to support weighted sampling of training datasets, configured through the yaml config file, e.g.,

```
stop_strategy: "last_exhausted"
data_sets: []
data_dicts:
    name: "clean"
    splits:
      - "train.100"
      - "train.360"
    weight: 2
    num_samples: 100_000
    name: "other"
    splits:
      - "train.500"
    weight: 1
    num_samples: 100_000
```

Three stop strategies are available: last_exhausted (default), first_exhausted, and never_stop. The first two strategies are similar to the logic used in the Hugging Face dataset, while the third option allows the maximum number of training steps to be determined by max_steps.

The training datasets are sampled based on their weights (default: 1.0) specified in the configuration file. In the example above, samples from the clean/other subsets are sampled at a 2:1 ratio.

### DIFF
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,6 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  # temperary for training verfication
-  UV_BRANCH: zhuang/add-generic-dataset
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -14,5 +14,5 @@ command: >-
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   # temperary for training verfication
-  UV_BRANCH: main
+  UV_BRANCH: zhuang/add-generic-dataset
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
-  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper.yaml
+  UV_BRANCH: zhuang/add-kl-loss
+  TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: zhuang/add-kl-loss
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
+  # Temporary branch for interleaving updates
+  UV_BRANCH: zhuang/update-interleaving
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,6 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  # temporary branch for running experiments
-  UV_BRANCH: zhuang/add-generic-dataset
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -3,7 +3,7 @@ name: ultravox
 image: mosaicml/composer:latest
 compute:
   gpus: 8
-  cluster: r7z22p1
+  cluster: r14z3p1
 integrations:
   - integration_type: git_repo
     git_repo: fixie-ai/ultravox
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -14,5 +14,5 @@ command: >-
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   # temperary for training verfication
-  UV_BRANCH: zhuang/add-kl-loss
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,6 +13,5 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  # Temporary branch for interleaving updates
-  UV_BRANCH: zhuang/update-interleaving
+  UV_BRANCH: main
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
+  # temperary for training verfication
+  UV_BRANCH: zhuang/add-kl-loss
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
+  # temporary branch for running experiments
+  UV_BRANCH: zhuang/add-generic-dataset
   TRAIN_ARGS: --config_path ultravox/training/configs/llama3_whisper_kd.yaml
--- a/ultravox/data/dataset_config.py
+++ b/ultravox/data/dataset_config.py
@@ -1,14 +1,7 @@
 import dataclasses
-from enum import Enum
 from typing import List, Optional
 
 from pydantic import BaseModel
-
-
-class InterleaveStopStrategy(str, Enum):
-    first_exhausted = "first_exhausted"
-    last_exhausted = "last_exhausted"
-    never_stop = "never_stop"
 
 
 class DataDictConfig(BaseModel):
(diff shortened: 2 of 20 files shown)


zqhuang211:
@juberti The following test unit failures need your attention.
```
======================================================================================================== short test summary info =========================================================================================================
ERROR inference/infer_test.py::test_infer_16kHz - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ERROR inference/infer_test.py::test_infer_48kHz - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ERROR inference/infer_test.py::test_infer_16kHz_stream - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
ERROR inference/infer_test.py::test_infer_text_only - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

zqhuang211:
> @juberti The following test unit failures need your attention.
> 
> ```
> ======================================================================================================== short test summary info =========================================================================================================
> ERROR inference/infer_test.py::test_infer_16kHz - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ERROR inference/infer_test.py::test_infer_48kHz - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ERROR inference/infer_test.py::test_infer_16kHz_stream - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ERROR inference/infer_test.py::test_infer_text_only - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ```

Never mind, it seems to be a version issue with my local copy. All checks passed on github.

mcloud.yaml
```diff
@@ -13,5 +13,6 @@ command: >-
   cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
-  UV_BRANCH: main
+  # Temporary branch for interleaving updates
+  UV_BRANCH: zhuang/update-interleaving
```
farzadab, line 17: seems pushed by mistake

ultravox/data/datasets.py
```diff
@@ -274,10 +275,15 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
         self._args = args
         self._session: Optional[requests.Session] = None
         self._rng = np.random.default_rng(self._args.shuffle_seed)
+        self._weight = 1.0  # the default weight for the dataset
```
farzadab, line 278: Hmm, having weight inside dataset seems a bit weird.

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
+        self._rng = np.random.default_rng(seed)
+        self._static = static
+
+        if stop_strategy not in ["first_exhausted", "last_exhausted", "never_stop"]:
+            raise ValueError(
+                "Invalid stop_strategy. Choose from 'first_exhausted', 'last_exhausted', 'never_stop'."
+            )
+        self._stop_strategy = stop_strategy
+
+        weights = [getattr(ds, "get_weight", lambda: 1)() for ds in datasets]
```
farzadab, line 1098: Multiple things don't seem to match up on this line. Are you sure it works correctly?
```suggestion
        weights = [getattr(ds, "weight", 1) for ds in datasets]
```

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
+        total_weight = sum(weights)
+        self._normalized_probs = [w / total_weight for w in weights]
 
     def __iter__(self):
+        # If no datasets are provided, return an empty iterator
+        if not self._datasets:
+            return
```
farzadab, line 1105: I don't have much experience with Python generators so just checking: are you sure returning `None` is the right thing here?

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
-                Let's say you have two datasets, A and B which have 5 and 3 samples respectively.
-
-                `repeat=False`: [A0, B0, A1, B1, A2, B2, A3, A4]
-                `repeat=True` : [A0, B0, A1, B1, A2, B2, A3, B0, A4, B1, A0, ...]
-
-                NOTE: with `repeat=True`, `__iter__` never stops.
+            datasets: A list of data.IterableDataset objects.
+            stop_strategy: Strategy for stopping iteration. Choose from 'first_exhausted', 'last_exhausted', 'never_stop'.
+            seed: Optional seed for reproducibility.
+            static: If true, the datasets are interleaved in a static order with equal weights.
```
farzadab, line 1085: It's not obvious what this means and why it's necessary. Why not just set all weights to 1 in this case?
The code is unnecessarily complicated because of this (IMO unnecessary) distinction.

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
-                else:
-                    iters[iter_index] = iter(self._datasets[iter_index])
+                # Reconstruct the iterator
+                iters[iter_index] = iter(self._datasets[iter_index])
+                exhausted[iter_index] = True
+                if self._static:
+                    static_iter = cycle(
+                        list(range(iter_index + 1, len(self._datasets)))
+                        + list(range(iter_index + 1))
+                    )
```
farzadab, line 1129: IIUC if dataset A is exhausted then dataset B, the dataset A gets added back into the mix.

It seems like a mix of the `repeat=False` and `repeat=True` behaviour. Which one did you want?

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
+        exhausted = [False] * len(iters)
+
+        if self._static:
+            static_iter = cycle(range(len(self._datasets)))
+
+        while True:
+            if self._static:
+                iter_index = next(static_iter)
+            else:
+                iter_index = self._rng.choice(len(iters), p=self._normalized_probs)
```
farzadab, line 1117: This totally ignores `exhausted`. Is that intentional?

ultravox/data/datasets.py
zqhuang211: Each dataset should have a weight (so that we can used weighted sampling), and the default is 1.0. In the GenericVoiceDataset, the default weight is 1.0 and can be configured in the config file. Currently, all other dataset classes share the same weight because these classes don’t support configurations. Once we update all dataset classes to support configuration via the YAML config, their weight can be configured individually (with the default weight being 1.0).

ultravox/data/datasets.py
zqhuang211: I think that signals when iterator ends, @juberti please double check

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,73 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
-    """Interleaves multiple IterableDataset objects."""
+    """Interleaves multiple IterableDataset objects based on normalized weights."""
 
     def __init__(
-        self, datasets: Sequence[data.IterableDataset], repeat: bool = False
+        self,
+        datasets: Sequence[data.IterableDataset],
+        stop_strategy: str = "last_exhausted",
+        seed: Optional[int] = 42,
+        static: bool = False,
```
juberti, line 1078: I don't think we need this for anything other than unit tests, right? Accordingly I think we should just use a fixed seed in the test.

ultravox/data/datasets.py
zqhuang211: When we do weighted sampling, we don't have to have `repeat=False`

ultravox/data/datasets.py
zqhuang211: I am not sure about this. Static removes random sampling and uses a static order, which could also be helpful for debugging, or for something we don't want any randomness, e.g., validation and evaluation.

ultravox/data/datasets_test.py
```diff
@@ -62,25 +62,31 @@ def test_dataproc():
     assert list(s) == [0, -1, -2, -3, -4]
```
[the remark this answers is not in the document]
zqhuang211: I added two additional tests.

ultravox/data/datasets.py
zqhuang211: exhausted in handled in the `except StopIteration` block

ultravox/data/datasets.py
juberti: yeah, since we `yield` in this generator method, exiting the method indicates that iteration has completed.

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,75 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
             except StopIteration:
-                if not self._repeat:
-                    iters.pop(iter_index)
-                    iter_index %= max(1, len(iters))
-                else:
-                    iters[iter_index] = iter(self._datasets[iter_index])
+                # Reconstruct the iterator
+                iters[iter_index] = iter(self._datasets[iter_index])
+                exhausted[iter_index] = True
+                if (
```
juberti, line 1121: There's no need for the `any` check here, right? We know we can bail out here.

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,75 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
-                yield val
+                yield next(iters[iter_index])
             except StopIteration:
-                if not self._repeat:
-                    iters.pop(iter_index)
-                    iter_index %= max(1, len(iters))
-                else:
-                    iters[iter_index] = iter(self._datasets[iter_index])
+                # Reconstruct the iterator
+                iters[iter_index] = iter(self._datasets[iter_index])
```
juberti, line 1119: We should only restart the iterator for `last_exhausted` and `never_stop`, I think.

ultravox/data/datasets.py
```diff
@@ -1060,43 +1068,75 @@ def create_dataset(name: str, args: VoiceDatasetArgs) -> data.IterableDataset:
...
         """
         super().__init__()
         self._datasets = datasets
-        self._repeat = repeat
+        self._rng = np.random.default_rng(seed)
+        self._static = static
+
+        self._stop_strategy = stop_strategy
+
+        weights = [getattr(ds, "weight", 1) for ds in datasets]
```
juberti, line 1094: Can we just access `ds.weight` directly here?

ultravox/data/dataset_config.py
```diff
@@ -1,16 +1,24 @@
 import dataclasses
+from enum import Enum
 from typing import List, Optional
 
 from pydantic import BaseModel
 
 
+class InterleaveStopStrategy(str, Enum):
```
juberti, line 8: I think I would keep this in datasets.py as it's specific to the `InterleaveDataset` class there. Also can the enum names be capitalized or does that cause a problem for config serialization?

ultravox/data/datasets.py
juberti: Understood, but in that case I tend to think it should be a separate class, rather than having a single class with two distinct code paths.

ultravox/data/datasets.py
zqhuang211: Will update to:

```
                exhausted[iter_index] = True

                # Check if stopping condition is met
                if (
                    self._stop_strategy
                    == dataset_config.InterleaveStopStrategy.first_exhausted
                    and any(exhausted)
                ):
                    break
                if (
                    self._stop_strategy
                    == dataset_config.InterleaveStopStrategy.last_exhausted
                    and all(exhausted)
                ):
                    break

                # Recreate the iterator if stopping condition is not met and yield the next sample
                yield next(iters[iter_index])
  ```

File: ultravox/data/datasets.py (hunk 2 above)
Comment (juberti) [reply ultravox/data/datasets.py:278]: FWIW, `datasets.interleave_datasets` passes in `probabilities` as an array of probabilities, allowing this logic to stay within the interleave method. I think that might turn out to be cleaner here, although it might require a bit more work to set up the config.

File: ultravox/data/datasets.py (hunk 11 above)
Comment (juberti) [reply ultravox/data/datasets.py:1119]: perhaps
```
if self._stop_strategy == StopStrategy.FIRST_EXHAUSTED or (
   self._stop_strategy == StopStrategy.LAST_EXHAUSTED and all(exhausted)):
   break
```

ultravox/data/datasets.py
zqhuang211: Even when weights are all 1, there is still randomness in sampling.

zqhuang211:
@juberti @farzadab Please check if all concerns have been addressed.

PR #71, opened 2024-08-06
Fix Model Name in README.md

msach22:


juberti:
Hi, the model is now known as v0_2 because HuggingFace doesn't seem to like dots in model names for some reason.

msach22:
Got it. Looks like the inference server needs to be updated because it responds with a 404 if using v0.2. (works for me, just an FYI)

PR #73, opened 2024-08-08
Add option to force push LLM weights

farzadab:
The vLLM code couldn't load the new [`ultravox-dev`](https://huggingface.co/fixie-ai/ultravox-dev) weights since the LLM weights weren't included. So to make things easier this PR allows for force pushing LLM weights.

A better way to solve this would've been to update vLLM, but for the time being I think this is a simpler and less bug-prone approach.

The model weights were also updated with the following command:
`poetry run python -m ultravox.tools.push_to_hub -m wandb://fixie/ultravox/model-zhuang.2024-07-15-ultravox.blsp-kd-4e:v5 -u fixie-ai/ultravox-dev --force_push_llm_weights`

### DIFF
--- a/ultravox/tools/push_to_hub.py
+++ b/ultravox/tools/push_to_hub.py
@@ -27,6 +27,8 @@ class UploadToHubArgs:
     data_type: Optional[str] = None
     # Public or private (default)
     private: bool = True
+    # Push LLM weights even if they are not fine-tuned
+    force_push_llm_weights: bool = True
 
 
 def main(args: UploadToHubArgs):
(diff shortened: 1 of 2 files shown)


ultravox/tools/push_to_hub.py
```diff
@@ -27,6 +27,8 @@ class UploadToHubArgs:
     data_type: Optional[str] = None
     # Public or private (default)
     private: bool = True
+    # Push LLM weights even if they are not fine-tuned
+    force_push_llm_weights: bool = True
```
juberti, line 31: Do we want to do this by default? I tend to think we should simply update our vllm fork to be smarter here rather than force all users to redownload the llama 3.1 weights.

juberti: To be clear, I'm OK to move forward with this PR, but I was thinking this option should default to `False`.

farzadab: I was holding off on the PR for now to see if @petersalas can solve the issue on his side. If it works, I'd rather remove the option tbh.

zqhuang211: This plan sounds good to me.

zqhuang211 (approved):

farzadab:
Peter has merged his changes to vllm, as such, this PR is not going to be needed. Closing.

PR #75, opened 2024-08-09
Update gradio demo to support text/voice conversation

zqhuang211:
Updated the Gradio demo to support both text and voice conversations. The implementation now caches past_key_values for previous conversation turns.

Updated `pyproject.toml` and `poetry.lock` to the latest versions of gradio (providing a more stable UI) and torch (adding mps support for Apple Silicon devices).

For machines with limited memory, use `--model_path wandb://fixie/ultravox/model-zhuang.2024-07-31-ultravox.blsp-kd-2b-tinyllama:v5`
For better results, use use `--model_path wandb://fixie/ultravox/model-zhuang.2024-07-31-ultravox.blsp-kd-2a:v5`


ultravox/inference/infer.py
```diff
@@ -11,7 +11,7 @@
 from ultravox.model import ultravox_processing
 
 SAMPLE_RATE = 16000
-MAX_TOKENS = 1024
+MAX_NEW_TOKENS = 1024
```
zqhuang211, line 14: Change `MAX_TOKENS` to `MAX_NEW_TOKENS` to be consistent with the parameter name `max_new_tokens` in `self.model.generate()`

ultravox/model/ultravox_model.py
```diff
@@ -247,10 +247,16 @@ def prepare_inputs_for_generation(
             **kwargs,
         )
 
-        if is_cache_empty(past_key_values) and audio_values is not None:
-            # We only want to use audio features in the 1st generation step
+        prefill_start_idx = kwargs["cache_position"][0]
```
zqhuang211, line 250: In order to support cached history in multi-turn conversation, past_key_value may not be None for the 1st generation step of a new turn.

ultravox/inference/infer.py
```diff
@@ -33,22 +33,33 @@ def __init__(
...
+        num_beams: int = 1,
     ) -> base.VoiceOutput:
         inputs = self._dataproc(sample)
         input_len = inputs["input_ids"].shape[1]
-        output = self._generate(inputs, max_tokens, temperature)
-        output_tokens = output[0][input_len:]
+        output = self._generate(
+            inputs, max_new_tokens, temperature, past_key_values, num_beams
+        )
+        output_tokens = output.sequences[0][input_len:]
```
zqhuang211, line 46: Due to passing `past_key_values` to self._generate(), the type of `output` changes from a list to a GenerateDecoderOnlyOutput. This causes failures in inference unit tests. @juberti Please take a look.

ultravox/inference/infer.py
```diff
@@ -33,22 +33,33 @@ def __init__(
     def infer(
         self,
         sample: datasets.VoiceSample,
-        max_tokens: Optional[int] = None,
+        max_new_tokens: Optional[int] = None,
         temperature: Optional[float] = None,
+        past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = None,
```
juberti, line 38: This parameter doesn't seem quite right to me since it has no meaning in other implementations. Instead could we just keep an internal cache within `LocalInference`?

ultravox/inference/infer.py
```diff
@@ -33,22 +33,33 @@ def __init__(
     def infer(
         self,
         sample: datasets.VoiceSample,
-        max_tokens: Optional[int] = None,
+        max_new_tokens: Optional[int] = None,
```
juberti, line 36: I think we want to keep the same parameter names as in the base class.

ultravox/inference/infer.py
juberti: FakeInference will probably need to be updated to return a `GenerateDecoderOnlyOutput` as well, should be a straightforward change.

ultravox/inference/base.py
```diff
@@ -45,8 +50,51 @@ def infer_stream(
...
+
+
+class History:
+    def __init__(self, audio_token_replacement: str = "<|eot_token|>"):
+        self.audio_token_replacement: str = audio_token_replacement
+        self.audio_placeholder = "<|audio|>"
+        self.messages: List[Dict[str, str]] = []
+        self.key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = None
+
+    def add_message(self, message: Dict[str, str], audio_token_len: int):
```
juberti, line 68: I'd prefer to keep the same interface here as in the model class, where we specify lists of audio insert position and audio len, rather than using a magic token string. Alternatively we could make the interface be a List of `MessagePart`, where each part is either a string or an audio chunk (similar to the OpenAI API)

Justfile
```diff
@@ -8,6 +8,7 @@ default: format check test
 
 install:
     pip install poetry==1.7.1
+    poetry config virtualenvs.in-project true --local
```
juberti, line 11: Do we need to make this change in this PR? I'd prefer to handle this separately if possible.

ultravox/inference/base.py
```diff
@@ -10,6 +12,8 @@ class VoiceOutput:
     text: str
     input_tokens: int
     output_tokens: int
+    audio_token_len: int = 0
```
juberti, line 15: maybe call this `input_audio_tokens` for consistency

Justfile
zqhuang211: Sure. This can be taken out.

ultravox/inference/base.py
zqhuang211: This field is no longer needed after adding history support inside LocalInference

ultravox/model/ultravox_model.py
farzadab: Are you absolutely sure `cache_positions` may never be None? Is that dependent on the `transformers` version?

Also if you need it, I suggest putting `cache_positions` in the function signature rather than getting it from `kwargs`. That way the type can also be specified.

ultravox/tools/infer_api.py
```diff
@@ -23,6 +24,7 @@ def infer(
         sample: datasets.VoiceSample,
         max_tokens: Optional[int] = None,
         temperature: Optional[float] = None,
+        past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = None,
```
farzadab, line 27: I'm confused why you need this. Even `LocalInference.infer` doesn't take `past_key_values`.

ultravox/model/ultravox_model.py
zqhuang211: `prepare_inputs_for_generation` is called in `transformers.generation.utils.GenerationMixin`  after `cache_position` is initialized:

`model_kwargs = self._get_initial_cache_position(input_ids, model_kwargs)`

zqhuang211: Good suggestion about putting `cache_position` in the function signature.

ultravox/tools/infer_api.py
zqhuang211: Good catch. `past_key_values` was removed from `LocalInference.infer` as we put `history` implementation inside `LocalInference`, so `past_key_values` is no longer needed.

ultravox/inference/infer.py
zqhuang211: Good suggestion. Updated to include `past_messages` and `past_key_values` within `LocalInference`

ultravox/inference/infer.py
farzadab: Yes I think it should change in base class and all the children.
Python never complains, so I guess keeping the interface consistent is on the developer :)

farzadab (approved):

ultravox/inference/infer.py
zqhuang211: I changed it back to `max_tokens` to be consistent with the base class (to be consistent with OPENAI parameter name)

ultravox/model/ultravox_model.py
```diff
@@ -237,20 +237,28 @@ def prepare_inputs_for_generation(
...
         **kwargs,
     ) -> Dict[str, Any]:
         model_input = self.language_model.prepare_inputs_for_generation(
             input_ids=input_ids,
             past_key_values=past_key_values,
             attention_mask=attention_mask,
             inputs_embeds=inputs_embeds,
+            cache_position=cache_position,
             **kwargs,
         )
```
liPatrick, line 251: How did this work before, if we were not offsetting the audio_token_start_idx by the prefill_start_idx?

zqhuang211: Previously we use `is_cache_empty(past_key_values)` to determine if we need to set the `audio` fields. The `audio` fields are only needed for the prefilling stage, which is the first generation step and `past_key_values=None`.

ultravox/tools/gradio_demo.py
```diff
@@ -1,42 +1,151 @@
...
 class DemoConfig:
     # model_path can refer to a HF hub model_id, a local path, or a Weights & Biases artifact
     #    fixie-ai/ultravox
     #    runs/llama2_asr_gigaspeech/checkpoint-1000/
     #    wandb://fixie/ultravox/model-llama2_asr_gigaspeech:v0
-    model_path: str = "fixie-ai/ultravox"
-    default_prompt: str = "Transcribe\n<|audio|>"
+    model_path: str = "fixie-ai/ultravox-v0_3"
+    # "mps", "cuda", or "cpu"
+    device: str = "mps"
```
juberti, line 25: These values (`device` and `data_type`) should probably be omitted because `UltravoxInference` will auto-select the best options automatically.

ultravox/inference/infer.py
```diff
@@ -24,24 +25,68 @@ def __init__(
...
+            audio_token_len = (
+                0 if "audio_token_len" not in inputs else inputs["audio_token_len"][0]
+            )
+            self._add_past_message(extended_sample.messages[-1], audio_token_len)
+            self._add_past_message({"role": "assistant", "content": output_text}, 0)
+            self.past_key_values = output.past_key_values
+
         return base.VoiceOutput(output_text, input_len, output_len)
 
     def infer_stream(
```
juberti, line 92: add a comment that supporting caching in stream mode is still a TODO.

ultravox/tools/gradio_demo.py
```diff
@@ -1,42 +1,151 @@
...
+    def process_audio(chatbot, prompt, audio, max_new_tokens, temperature):
+        return process_turn(
+            chatbot,
+            prompt,
+            audio=audio,
+            max_new_tokens=max_new_tokens,
+            temperature=temperature,
+        )
+
+    def gradio_reset():
```
juberti, line 92: This feels fragile, i.e., if the user reloads the page the conversation won't get reset. This makes me think that the actual chat history should be kept here in the chatbot, but we can keep the KV cache in `LocalInference`. This is similar to how the production (OpenAI, etc) services work.

If you think this is too complex to take on here I'd be OK with fixing this in a followup PR.

juberti (approved):

ultravox/tools/gradio_demo.py
zqhuang211: I will update `ultravox.inference.utils.default_device()` to include `mps`. But it is still good to keep the options here (default `None`), e.g., testing `cpu` speed on Macbook.

ultravox/tools/gradio_demo.py
zqhuang211: This is fixed by adding one line:
```
demo.load(gradio_reset, [], [chatbot, prompt, audio], queue=False)
```

ultravox/inference/infer.py
```diff
@@ -24,24 +25,68 @@ def __init__(
...
+        self.past_messages: List[Dict[str, str]] = []
+        self.past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = (
+            None
+        )
+
+    def reset_conversation(self):
+        self.past_messages = []
+        self.past_key_values = None
+
+    def _add_past_message(self, message: Dict[str, str], audio_token_len: int):
```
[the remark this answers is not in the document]
zqhuang211: Good suggestion. Changed it to:

```
    def update_conversation(
        self,
        past_messages: List[Dict[str, str]] = [],
        past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = None,
    ):
        self.past_messages = past_messages
        self.past_key_values = past_key_values
 ```

PR #81, opened 2024-08-14
Add retry in ds tool

liPatrick:


ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -134,6 +152,10 @@ def _map_sample(self, sample, exclude_fields):
             rendered = jinja2.Template(
                 self.template, undefined=jinja2.StrictUndefined
             ).render(**filtered_sample, json_dump=json.dumps, text_proc=text_proc)
+            if rendered.strip() == "":
+                sample[self.new_column_name] = ""
```
juberti, line 156: If you want to reject the sample entirely, maybe just return `None` here to make the subsequent filter trivial (eg `lambda sample: sample is not None`).

juberti: also, is it worth logging here?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -230,6 +253,10 @@ def main(args: DatasetToolArgs):
         data_dict[split] = args.task.map_split(
             ds_split, args.num_workers, args.writer_batch_size, args.exclude_fields
         )
+        if isinstance(args.task, TextGenerationTask):
```
juberti, line 256: Suggest adding a 1-line comment explaining the intent here.

juberti (approved):
LG, you might want to generalize the `filter` handling as noted below.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Ah I realized I had made a mistake here, and its actually supposed to check the output of textproc(sample.text), not rendered. Rendered is never going to be empty here because there's the other template text. I made a change to pass in the column we want to filter in the args instead.

ultravox/data/text_proc.py
```diff
@@ -30,6 +30,8 @@ def format_asr_text(text: str) -> str:
     """
     remaining_words = []
     for word in text.split():
+        if word in GIGASPEECH_GARBAGE_UTTERANCE_TAGS:
```
juberti, line 33: Do we want to throw away the entire utterance here, or just the <NOISE>?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -230,6 +255,14 @@ def main(args: DatasetToolArgs):
         data_dict[split] = args.task.map_split(
             ds_split, args.num_workers, args.writer_batch_size, args.exclude_fields
         )
+        # Filter out samples where column_to_check is empty after text_proc.format_asr_text
+        if isinstance(args.task, TextGenerationTask) and args.column_to_check_filter:
```
juberti, line 259: I'd probably push this into `TextGenerationTask.map_split` to encapsulate this logic in one place. Also, will `column_to_check-filter` ever be different than `new_column_name`? ie could you just check the output of the chat completion?

ultravox/data/text_proc.py
liPatrick: I think for now we throw away the whole utterance, but its subject to change depending on how we want to run our experiments.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: column_to_check_filter will always be different than new_column_name. The idea is if the text we pass into chat completion is empty, our chat completion will respond with something like "No text has been provided... etc", which is a row we shoudl throw out.

juberti: ah ok I see what you're doing now. Maybe the best option here is to make the template raise an error if the data is ill-formed, e.g., https://stackoverflow.com/questions/21778252/how-to-raise-an-exception-in-a-jinja2-macro? Then you can push this logic entirely into the template.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -47,6 +47,7 @@ def map_split(
         num_proc: int,
         writer_batch_size: int,
         exclude_fields: List[str],
+        include_filter: bool = False,
```
liPatrick, line 50: Adding this flag here because for some datasets we don't expect there to be any empty text values, and for those we don't want to run through an unncessary filter map since its pretty time consuming.

juberti: Might also be able to just count the number of template failures and only filter when != 0.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -115,32 +117,46 @@ def map_split(
...
             lambda sample: self._map_sample(sample, set(exclude_fields)),
             num_proc=num_proc,
             writer_batch_size=writer_batch_size,
         )
+        if self.template_failures == 0:
+            return ds_mapped
+
+        # Filter out samples where new_column_name is None
+        return ds_mapped.filter(
+            lambda sample: (True if sample[self.new_column_name] != None else False),
```
juberti, line 130: could be written as `sample[self.new_column_name] != None`

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -115,32 +117,46 @@ def map_split(
...
+
         except jinja2.TemplateError as e:
-            print(f"Error rendering template: {e}")
-            print(f"template: {self.template}")
-            print(f"sample keys: {list(filtered_sample.keys())}")
-            raise ValueError(
-                f"Template rendering failed. Make sure all keys in the template exist in the sample."
-            ) from e
+            error_message = str(e)
+            if "The formatted text is empty." in error_message:
```
juberti, line 149: Glad to see this worked out, but this check seems a bit fragile. Is there no way to distinguish these two cases by the exception type?

juberti (commented):
a couple nits for the future

Issue #68, opened 2024-08-05
Support offline batch inference mode


zqhuang211:
Currently, the inference module only supports processing one sample at a time. It needs to be updated to support batch mode for both synthetic data generation (e.g., continuation) and after evaluation.

juberti:
Fixes in #82

PR #82, opened 2024-08-14
Offline batch inference mode

liPatrick:
Adds batching capabilities. 
addresses issue: https://github.com/fixie-ai/ultravox/issues/68


ultravox/inference/base.py
```diff
@@ -40,6 +40,15 @@ def infer(
...
         pass
 
+    def batch_infer(
+        self,
+        samples: List[datasets.VoiceSample],
+        batch_size: int = 1,
+        max_tokens: Optional[int] = None,
+        temperature: Optional[float] = None,
+    ):
+        return self.batch_infer(samples, batch_size, max_tokens, temperature)
```
farzadab, line 50: I'm confused, what does this function do? Did you want it to be abstract and/or to call `infer` repeatedly instead?

ultravox/inference/infer.py
```diff
@@ -68,7 +124,8 @@ def infer_stream(
         yield base.InferenceStats(input_tokens, output_tokens)
         thread.join()
 
-    def _dataproc(self, sample: datasets.VoiceSample):
+    def _dataproc(self, sample: datasets.VoiceSample, batch=False):
+        print("sample", sample)
```
farzadab, line 128: nit: remove print

ultravox/inference/infer.py
```diff
@@ -102,6 +159,9 @@ def _dataproc(self, sample: datasets.VoiceSample):
         inputs = {k: v.to(self.model.device) for k, v in inputs.items()}
         if "audio_values" in inputs:
             inputs["audio_values"] = inputs["audio_values"].to(dtype=self.dtype)
+        if batch:
+            for key, val in inputs.items():
+                inputs[key] = val.squeeze()
```
farzadab, line 164: nit: are you sure you want `squeeze()` or `squeeze(0)`?
Asking because the behaviour of `squeeze` is dynamic and can remove too many columns depending on the values.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,44 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+            question_text = sample.audio_transcript
+            expected_answer = sample.messages[-1]["content"]
+            # Drop any assistant response from the sample.
+            sample.messages = sample.messages[:-1]
+            if not args.json:
+                run_tui(i, inference, sample, args, expected_answer, scores)
+            else:
+                output = inference.infer(
+                    sample, max_tokens=args.max_tokens, temperature=args.temperature
+                )
```
farzadab, line 226: I suggest we combine `batch` mode and `json` mode together if you can. Now there are 3 paths for the code to take which seems like too many.

farzadab: Using [BatchSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.BatchSampler) might help.

ultravox/data/datasets.py
```diff
@@ -76,29 +76,27 @@ class DataCollatorForSeq2SeqWithAudio(transformers.DataCollatorForSeq2Seq):
...
-        if self.include_alt_fields:
-            # these fields are hard-coded in the transformer data collator, so they need special handling before calling the super method
-            alt_features = [
-                {
-                    "input_ids": f.pop("alt_input_ids"),
-                    "attention_mask": f.pop("alt_attention_mask"),
-                    "labels": f.pop("alt_labels"),
-                }
-                for f in features
-            ]
```
farzadab, line 88: The reason I had removed this stuff was because it was never needed in inference mode. We shouldn't be removing it in here though, else KD training won't work.

farzadab (commented):
Thanks Pat. Not sure if this PR was ready or not, but I left some comments.

ultravox/inference/base.py
liPatrick: Yeah, I'm just going to add a type check in infer_tool.dataset_infer, and i can get rid of this function.

ultravox/inference/infer.py
liPatrick: right, squeeze(0) is definitly correct here, thanks for the catch.

ultravox/data/datasets.py
```diff
@@ -76,6 +76,12 @@ class DataCollatorForSeq2SeqWithAudio(transformers.DataCollatorForSeq2Seq):
 
     def __call__(self, features, *args, **kwargs):
         audio_values = [f.pop("audio_values", None) for f in features]
+        for i, audio in enumerate(audio_values):
+            audio_values[i] = torch.tensor(audio)
+
+        for i, feature in enumerate(features):
+            features[i]["input_ids"] = torch.tensor(feature["input_ids"])
+
```
zqhuang211, line 84: Can you explain why the changes in this method is needed for batched inference?

ultravox/inference/base.py
```diff
@@ -40,6 +40,15 @@ def infer(
...
 
+    # def batch_infer(
+    #     self,
+    #     samples: List[datasets.VoiceSample],
+    #     batch_size: int = 1,
+    #     max_tokens: Optional[int] = None,
+    #     temperature: Optional[float] = None,
+    # ):
+    #     return self.batch_infer(samples, batch_size, max_tokens, temperature)
+
```
zqhuang211, line 51: Remove the commented lines?

ultravox/inference/infer.py
```diff
@@ -30,6 +32,60 @@ def __init__(
         self.processor = processor
         self.dtype = dtype
 
+    def batch_infer(
+        self,
+        all_samples: List[datasets.VoiceSample],
+        batch_size: int = 1,
+        max_tokens: Optional[int] = None,
+        temperature: Optional[float] = None,
+    ) -> base.InferenceGenerator:
```
zqhuang211, line 41: If `all_samples` is a list of `datasets.VoiceSample`, it means all audio files need to be loaded in memory. It will not scale for large datasets.

ultravox/inference/infer.py
```diff
@@ -30,6 +32,60 @@ def __init__(
...
+                    "index": i,
+                    "question_text": question_text,
+                    "expected_answer": expected_answer,
+                }
+            )
+            sample.messages = sample.messages[:-1]
+
+        dataset: Dataset[Any] = Dataset.from_list(
+            [self._dataproc(s, batch=True) for s in all_samples]
+        )
```
zqhuang211, line 61: Similarly, applying `self._dataproc` to all samples at once will not scale for a large data set. Take a look at  `train.py` on how data processor in used in datasets and in batching. I think data processor should be applied to one batch at a time.

ultravox/inference/infer.py
liPatrick: Ah yes, i forgot about this, i need to make sure not to use lists anywhere when dealing with the datasets. Its tricky because I originally used the .map funciton but that seemed to corrupt the data it was mapping.

ultravox/inference/infer.py
```diff
@@ -30,6 +32,60 @@ def __init__(
...
+        temperature: Optional[float] = None,
+    ) -> base.InferenceGenerator:
+        data_collator = datasets.DataCollatorForSeq2SeqWithAudio(
+            tokenizer=self.tokenizer,
+            include_alt_fields=False,
+        )
+        sample_additional_info = []
+        for i, sample in enumerate(all_samples):
+            question_text = sample.audio_transcript
+            expected_answer = sample.messages[-1]["content"]
```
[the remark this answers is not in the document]
zqhuang211: The naming of `question_text` is also problematic. A sample doesn't necessarily have `audio` or `audio_transcript` in inference mode.

ultravox/data/datasets.py
farzadab: In training we only use right padding. If `left` padding is applied, all the token positions are gonna move and that includes where the audio is located which we assume is indicated by `audio_token_start_idx`. As such, we need to update this variable to match the reality.

liPatrick:
Sorry for the few iterations. Ended up working out a batch implementation that doesn't use dataloader. This was because it was tricky to use dataloader effectively without introducing awkwardness in the dataprocessing in infer_tools, and function definitions for batch_infer.

ultravox/inference/infer.py
```diff
@@ -38,6 +38,27 @@ def __init__(
             None
         )
 
+    def batch_infer(
+        self,
+        samples: List[datasets.VoiceSample],
+        max_tokens: Optional[int] = None,
+        temperature: Optional[float] = None,
+    ) -> base.InferenceGenerator:
+        inputs = [self._dataproc(s, batch=True) for s in samples]
```
farzadab, line 47: You need to enforce `tokenizer.padding_side = "left"` somewhere, most likely in the `__init__` function.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +191,47 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
-        if not args.json:
+
+    if args.json and isinstance(inference, infer.LocalInference):
+        if not args.batch_size:
+            args.batch_size = 1
+
+        start_time = time.time()
+        current_batch = []
+        for i, sample in enumerate(datasets.Range(ds, args.num_samples)):
+            current_batch.append(sample)
```
farzadab, line 202: An issue with creating the batch manually (as opposed to using a PyTorch `DataLoader`) is that everything will happen on a single thread and that your GPUs will be waiting for the data to get ready.
The `DataLoader` on the other hand uses thread(s) in the background to make sure the data is always needed when the GPU needs it.

I'm okay with not making the change right now (since that might make keeping `expected_answer` harder), but let's at least put a TODO or comment here that explains that there is room for improvement.

farzadab:
Thanks Patrick, looks good. Just a few minor comments.

ultravox/inference/infer.py
liPatrick: For my understanding, is this only necessary for batch infer? Or should we have done this for regular infer as well.

ultravox/tools/infer_tool.py
zqhuang211: @farzadab is there a way to use DataLoader without using a data_collator (or using a dummy data_collator), or some other type of DataLoader that uses multithreads for loading data?

ultravox/inference/infer.py
farzadab: Correct. For regular infer, there isn't any padding being applied anywhere, so it doesn't matter.

For batched generation however, the sizes of the input_ids will likely differ, hence the need for padding.
We want to be able to add new (generated) tokens on the right, and that won't work correctly if we've already padded the input on the right side.

ultravox/tools/infer_tool.py
```diff
@@ -77,6 +76,8 @@ class InferArgs:
     verbose: bool = simple_parsing.field(default=False, alias="-v")
     # JSON output
     json: bool = simple_parsing.field(default=False)
+    # Batch size
+    batch_size: Optional[int] = simple_parsing.field(default=None, alias="-b")
```
zqhuang211, line 80: should `batch_size=1` by default?

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +191,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+        if not args.batch_size:
+            args.batch_size = 1
+
+        # TODO: Add multithreading support for preparing the batch.
+        start_time = time.time()
+        current_batch = []
+        for i, sample in enumerate(datasets.Range(ds, args.num_samples)):
+            current_batch.append(sample)
+            if len(current_batch) == args.batch_size:
+                output = []
```
zqhuang211, line 205: What happens if the last batch doesn't have `args.batch_size` samples?

liPatrick: Yes, good catch it will skip it. I'll add a fix

ultravox/inference/infer.py
liPatrick: Ah thats right, thanks for clarifying

ultravox/inference/infer.py
```diff
@@ -38,6 +38,29 @@ def __init__(
...
+        data_collator = datasets.DataCollatorForSeq2SeqWithAudio(
+            tokenizer=self.tokenizer,
+            include_alt_fields=False,
```
zqhuang211, line 52: should the `data_collator` be a member of the inference instance, given that it is needed for each batch?

ultravox/inference/infer.py
```diff
@@ -38,6 +38,29 @@ def __init__(
...
         )
 
+        assert self.tokenizer.padding_side == "left"
+
+    def batch_infer(
+        self,
+        samples: List[datasets.VoiceSample],
+        max_tokens: Optional[int] = None,
+        temperature: Optional[float] = None,
+    ) -> base.InferenceGenerator:
```
zqhuang211, line 48: If the input is a list of `VoiceSample`, why not return a list of `VoiceOutput` as output? It seems conceptually more straightforward.

ultravox/inference/infer.py
```diff
@@ -152,6 +175,9 @@ def _dataproc(self, sample: datasets.VoiceSample):
         inputs = {k: v.to(self.model.device) for k, v in inputs.items()}
         if "audio_values" in inputs:
             inputs["audio_values"] = inputs["audio_values"].to(dtype=self.dtype)
+        if batch:
+            for key, val in inputs.items():
+                inputs[key] = val.squeeze(0)
         return inputs
```
zqhuang211, line 181: How does it compare to always assume `batch=True` or `batch=False`, i.e.., removing the `batch` parameter, and update the `infer()`, `infer_stream()`, `infer_batch()` methods accordingly? Logically, the output of `_dataproc` is processed by `_generate()`

ultravox/tools/infer_tool.py
farzadab: You can just use a pass-through `collate_fn`, e.g. `torch.utils.data.DataLoader(ds, batch_size=4, collate_fn=lambda x: x)`:

```python
from ultravox.data import datasets
import torch.utils.data
ds = datasets.LibriSpeechDummyDataset(datasets.VoiceDatasetArgs())
dl = torch.utils.data.DataLoader(ds, batch_size=4, collate_fn=lambda x: x)
print(next(iter(dl)))
```

Outputs:
```
[VoiceSample(messages=[{'role': 'user', 'content': 'Transcribe\n<|audio|>'}, {'role': 'assistant', 'content': 'Mister Quilter is the apostle of the middle classes and we are glad to welcome his gospel'}], audio=array([...
...
]
```

ultravox/inference/infer.py
liPatrick: Yeah I think this will probably be cleaner, because the squeeze(0) is more tightly coupled with the input into collator as opposed to what data_proc is supposed to output.

ultravox/inference/infer.py
liPatrick: Yeah, would be nice to not reinitalize everytime we do a batch.

ultravox/inference/infer.py
liPatrick: Although, this adds another loop through the already processed inputs. What do you think about the new change?

ultravox/tools/infer_tool.py
liPatrick: I've added in the dataloader now

ultravox/tools/infer_tool.py
```diff
@@ -1,19 +1,19 @@
...
 import argparse
 import dataclasses
-import json
 import os
 import time
 from typing import IO, List, Optional
 
 import numpy as np
 import simple_parsing
+from torch.utils.data import DataLoader
```
juberti, line 10:
```suggestion
from torch.utils import data as data_utils
```

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
-        expected_answer = sample.messages[-1]["content"]
-        # Drop any assistant response from the sample.
-        sample.messages = sample.messages[:-1]
-        if not args.json:
-            run_tui(i, inference, sample, args, expected_answer, scores)
-        else:
-            output = inference.infer(
-                sample, max_tokens=args.max_tokens, temperature=args.temperature
+
+    if args.json and isinstance(inference, infer.LocalInference):
```
juberti, line 196: I tend to think we should degrade gracefully here and just simulate `infer_batch` when an optimized batch API doesn't exist, via looping. The actual looping implementation could be in the base class.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
-            print(json.dumps(obj))
+            for i, output_text in enumerate(output_batch):
+                output[i]["output_text"] = output_text
+            print(output)
+        print("Total time", time.time() - start_time)
+
+    else:
+        scores: List[float] = []
+        for i, sample in enumerate(datasets.Range(ds, args.num_samples)):
+            # Store the original question and answer for JSON output.
```
juberti, line 231: question_text isn't needed any longer and this comment is out of date.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+            output = []
+            for sample in current_batch:
+                output.append(
+                    {
+                        "index": sample_index,
+                        "question_text": sample.audio_transcript,
+                        "expected_answer": sample.messages[-1]["content"],
+                    }
+                )
+                sample.messages = sample.messages[:-1]
```
zqhuang211, line 215: How do we handle the cases where the dataset doesn't have reference for the samples?

ultravox/tools/infer_tool.py
juberti: Otherwise we'll malfunction if args.json is set and a different impl is used.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
-            obj = {
-                "question": question_text,
-                "generated_answer": output.text,
-                "expected_answer": expected_answer,
-            }
-            print(json.dumps(obj))
+            for i, output_text in enumerate(output_batch):
+                output[i]["output_text"] = output_text
+            print(output)
+        print("Total time", time.time() - start_time)
```
juberti, line 226: I don't think we want to mix JSON and freeform text.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+                max_tokens=args.max_tokens,
+                temperature=args.temperature,
             )
-            obj = {
-                "question": question_text,
-                "generated_answer": output.text,
-                "expected_answer": expected_answer,
-            }
-            print(json.dumps(obj))
+            for i, output_text in enumerate(output_batch):
```
zqhuang211, line 223: Maybe add a check to make sure `len(output_batch) = len(output)`?

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+        )
+        current_batch = []
+        sample_index = 0
+        for current_batch in dl:
+            output = []
+            for sample in current_batch:
+                output.append(
+                    {
+                        "index": sample_index,
+                        "question_text": sample.audio_transcript,
```
juberti, line 211: Is the change to the JSON format intentional?

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,48 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
             )
-            obj = {
-                "question": question_text,
-                "generated_answer": output.text,
-                "expected_answer": expected_answer,
-            }
-            print(json.dumps(obj))
+            for i, output_text in enumerate(output_batch):
+                output[i]["output_text"] = output_text
+            print(output)
```
juberti, line 225: Is a `json.dumps` needed here? I assume the output is jsonl.

ultravox/inference/infer.py
```diff
@@ -37,6 +37,37 @@ def __init__(
...
             None
         )
+        self.data_collator = datasets.DataCollatorForSeq2SeqWithAudio(
+            tokenizer=self.tokenizer,
+            include_alt_fields=False,
+        )
+
+        assert self.tokenizer.padding_side == "left"
+
+    def batch_infer(
```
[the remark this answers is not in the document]
liPatrick: I think that will involve changing the return signature of infer, which we can do, but a lot of other functions depend on infer, such as all the functions in infer_api. I can do it if you think it makes sense to make all those downstream changes?

ultravox/tools/infer_tool.py
liPatrick: Yes, I would like to follow the same format as in regular infer. Although, I'm just realizing the keys don't match so I'll update that.

ultravox/tools/infer_tool.py
liPatrick: Why is it bad to mix the two? Aren't they just logs at the end of the day?

ultravox/tools/infer_tool.py
liPatrick: Ah the code is sort of hardcoded to expect that right now. Do think we should make that change _after_ we start seeing datasets that don't have reference in the samples? Else I'd imagine we could add a flag for that and have another line of logic to handle the other case.

ultravox/tools/infer_tool.py
liPatrick: For my own understanding, why is it bad to mix JSON and freeform text? Aren't they just logs at the end of the day?

juberti: a JSONL parser will explode if it gets non-JSON text, right?

ultravox/inference/infer.py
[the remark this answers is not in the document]
juberti: Understood, let's just do the `infer_batch` rename and we can revisit this in a followup. I'm doing some nearby changes so it's going to be a bit of a moving target anyway.

ultravox/tools/infer_tool.py
liPatrick: Is it common to use a jsonl parser for logs? I thought logs are mostly for debugging purpose, and not meant for further post processing

ultravox/tools/infer_tool.py
zqhuang211: Sounds good. I can imagine two use cases.

1. generating synthesized data for model training
2. self-evaluation: comparing SLM's output against LLM's output, where no reference is available.

Let's try to merge this PR first.

ultravox/inference/infer.py
```diff
@@ -37,6 +37,12 @@ def __init__(
         self.past_key_values: Optional[Union[Tuple, transformers.cache_utils.Cache]] = (
             None
         )
+        self.data_collator = datasets.DataCollatorForSeq2SeqWithAudio(
+            tokenizer=self.tokenizer,
+            include_alt_fields=False,
+        )
+
+        assert self.tokenizer.padding_side == "left"
```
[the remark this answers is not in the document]
liPatrick: Its already set in UltravxoInference, line 48.

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,46 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
+            for sample in current_batch:
+                output.append(
+                    {
+                        "index": sample_index,
+                        "question": sample.audio_transcript,
+                        "expected_answer": sample.messages[-1]["content"],
+                    }
+                )
+                sample.messages = sample.messages[:-1]
+                sample_index += 1
```
farzadab, line 215: A total nit, but I felt like this would be a bit cleaner (see next comment for related change):
```suggestion
            expected_answers = [sample.messages.pop()["content"] for sample in current_batch]
```

ultravox/tools/infer_tool.py
```diff
@@ -190,25 +192,46 @@ def dataset_infer(inference: base.VoiceInference, args: InferArgs):
...
-            obj = {
-                "question": question_text,
-                "generated_answer": output.text,
-                "expected_answer": expected_answer,
-            }
-            print(json.dumps(obj))
+            assert len(output) == len(output_batch)
+            for i, output_val in enumerate(output_batch):
+                output[i]["generated_answer"] = output_val.text
+            print(json.dumps(output))
```
farzadab, line 225: The `output` is a list, which is wrong I believe: each sample is assumed to be outputted on a different line.

Also here's a nitpicking proposal (needs to be combined with other change):
```suggestion
            assert len(current_batch) == len(output_batch)
            for sample, generated, expected  in zip(current_batch, output_batch, expected_answers):
                output = {
                    "index": sample_index,
                    "question": sample_in.audio_transcript,
                    "expected_answer": expected,
                    "generated_answer": generated.text,
                }
                sample_index += 1
                print(json.dumps(output))
```

farzadab (approved):
Looking much better after the other reviews!

I added some comments, but other than padding left, the others are mostly nit, so it's approved IMO.

ultravox/inference/infer.py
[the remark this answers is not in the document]
farzadab: Oh we had it already. Got it.

ultravox/tools/infer_tool.py
liPatrick: Thanks! Much cleaner :)

PR #85, opened 2024-08-15
Integrating OAI evals post training

farzadab:
This PR switches our end of training evaluations to use the OpenAI evals framework.


ultravox/training/train.py
```diff
@@ -78,6 +81,19 @@ def main() -> None:
...
+    local_rank = int(os.environ.get("LOCAL_RANK", 0))
+    is_master = local_rank == 0
+
+    if args.do_train:
+        train(args)
+
+    if args.do_eval and is_master:
+        gc.collect()
+        torch.cuda.empty_cache()
+        evaluate(args)
```
farzadab, line 93: The separation was the best way I knew to make sure all GPU memory is released for oaieval to take over.

ultravox/training/train.py
```diff
@@ -282,48 +297,103 @@ def main() -> None:
...
+def run_oaievalset(
+    log_dir: str, model_dir: str, eval_set: str, num_samples: Optional[int] = None
+):
+    env = os.environ.copy()
+
+    # num_gpus = max(1, torch.cuda.device_count())
+    env["EVALS_THREADS"] = "64"
+
+    # TODO: currently running this on a single GPU is faster than multiple GPUs :facepalm:
+    env["CUDA_VISIBLE_DEVICES"] = "0"
```
farzadab, line 374: After more debugging it turns out that:
1. the fix to `BatchedProcessPoolExecutor` did not help, but make inference use mainly `batch_size=1` which the following PR fixes: https://github.com/fixie-ai/evals/pull/13/files
2. Multi-GPU was somehow slower than using 8 GPUs. I'm not sure why but I couldn't figure it out easily so I think I'm gonna merge this first and then go for optimizing the 8 GPU inference.

ultravox/training/train.py
```diff
@@ -282,48 +297,103 @@ def main() -> None:
...
+    command = [
+        "oaievalset",
+        "--record_dir",
+        log_dir,
+        "generation/gpu/ultravox-dev",
+        eval_set,
+        f"--completion_args=model={model_dir}",
+    ]
+    if num_samples:
+        command.append(f"--max_samples={num_samples}")
```
farzadab, line 385: Only doing inference on the first 1024 samples rather than the whole set since evaluation is not as fast as I wanted. If we don't care about the speed, we can do the full set too @zqhuang211.

farzadab:
Here's an example task which barely does any training, but does the evaluation (assuming [this PR is merged](https://github.com/fixie-ai/evals/pull/13/files)): https://wandb.ai/fixie/ultravox/runs/nxgfzg3v

ultravox/training/train.py
```diff
@@ -78,6 +80,19 @@ def main() -> None:
...
+    if args.do_eval and is_master:
+        gc.collect()
+        torch.cuda.empty_cache()
+        evaluate(args)
+
+
+def train(args: config_base.TrainConfig):
     world_size = int(os.environ.get("WORLD_SIZE", 1))
     local_rank = int(os.environ.get("LOCAL_RANK", 0))
     is_master = local_rank == 0
```
zqhuang211, line 98: `is_master` is also computed earlier on line 84, maybe pass it in as an argument?

ultravox/training/train.py
```diff
@@ -282,48 +297,103 @@ def main() -> None:
...
-            dtype=dtype,
+
+    # Training loop
+    logging.info("Starting training...")
+    t_start = datetime.now()
+    logging.info(f"train start time: {t_start}")
+    if args.val_steps:
+        trainer.evaluate()
+    trainer.train()
+    if trainer.is_world_process_zero():
```
zqhuang211, line 308: is this different from `if is_master:`

ultravox/training/train.py
```diff
@@ -282,48 +297,103 @@ def main() -> None:
...
-        logging.info(f"eval end time: {t_end}")
-        logging.info(f"elapsed: {t_end - t_start}")
+    logs_dir = wandb.run.dir if wandb.run else str(args.logs_dir)
+
+    # Run audio-based evaluations and log to W&B
+    audio_metrics = run_oaievalset(
+        log_dir=os.path.join(logs_dir, "oaieval/audio"),
+        model_dir=str(args.output_dir),
+        eval_set="audio-core",
+        num_samples=args.eval_num_samples,
```
zqhuang211, line 341: Ideally, `eval_num_samples` should be configurable for each eval dataset separately. We can address this in another PR.

ultravox/model/ultravox_model.py
```diff
@@ -9,13 +9,13 @@
...
 import transformers.modeling_outputs
 import transformers.models
+from transformers.models.whisper import modeling_whisper as whisper
 
 # We must use relative import in this directory to allow uploading to HF Hub
 # Even "from . import X" pattern doesn't work (undocumented and unclear why)
 from .ultravox_config import LossConfig
 from .ultravox_config import LossFunction
 from .ultravox_config import UltravoxConfig
-from .whisper_model_modified import WhisperEncoder as ModifiedWhisperEncoder
```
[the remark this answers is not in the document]
farzadab: I hate mashing everything into one file too, but whenever I tried to load the model from disk with HF's `trust_remote_code=True`, it only loads the main files and throws an error saying `whisper_model_modified` does not exist.

ultravox/training/train.py
farzadab: It didn't seem like a big deal to keep it. Adding 3 new parameters to the function signature on the other hand didn't seem pretty.

ultravox/training/train.py
farzadab: In one node no. In multiple nodes, it enforce both that local_rank is zero and that node_rank is also zero, hence the word "world" process zero.

I can revert to using `is_master` for now.

ultravox/training/train.py
farzadab: Yeah originally I wanted to run the full set even, but with the multiple GPU issue I ended up restricting it again so that runs wouldn't take too long (though the default num_samples has already been increased by 4x). Hence I view this as a more of a temporary thing.

ultravox/training/train.py
zqhuang211: Understood. Does it require updates to the usage of `is_master` (e.g., eval) in other parts of the code when operating in multiple nodes?

ultravox/training/train.py
zqhuang211: I see, all three variables are used in multiple places in `train`.

ultravox/training/train.py
zqhuang211: Then maybe use `is_master` for now.

ultravox/training/train.py
```diff
@@ -282,48 +297,103 @@ def main() -> None:
...
+        f"--completion_args=model={model_dir}",
+    ]
+    if num_samples:
+        command.append(f"--max_samples={num_samples}")
+
+    # Run the evaluation set
+    subprocess.run(command, check=True, env=env)
+
+    # Extract the results from the log directory
+    df = make_table.extract_results(pathlib.Path(log_dir))
```
juberti, line 391: It feels somewhat hacky to have this sort of implicit dependency. Maybe we could adjust `make_table.py` to write out a dataframe that we could load, rather than us having to actually call into that module directly?

juberti (approved):
Overall this LGTM, nice work.

pyproject.toml
```diff
@@ -8,8 +8,8 @@ license = "MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.11"
-torch = "2.2.2"
+python = ">=3.10,<4.0"
```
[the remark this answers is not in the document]
juberti: seems fine

juberti:
Not required for this PR, but it looks like wandb has some basic support for handling oaieval outputs, including sample-by-sample visualization, maybe if we can log to the right format we can get this visualization too: https://wandb.ai/wandb_fc/openai-evals/reports/OpenAI-Evals-Demo-Using-W-B-Prompts-to-Run-Evaluations--Vmlldzo0MTI4ODA3

juberti:
example wandb oaieval run: https://wandb.ai/wandb/jobs/runs/ugqqpjff?nw=nwuser_scott

ultravox/training/train.py
farzadab: Hmm I hear you and I did the change, but I still don't see the difference in the interface or dependency: we still need to call exactly the same module, and know where it stores the final file, and then load it, knowing that it's a CSV (instead of PD).

Both the inputs and outputs are the same. The only difference is that we're now calling it through the shell instead of Python which is worse :?

I think the right approach is to pull everything into the Ultravox repo TBH. I'm gonna do that later on, but not today.

farzadab:
Justin using `wandb.Table` was a great idea. It shows all metrics in one table and also allows us to automatically create charts from them.
The charts from multiple experiments can also be stacked, even if they only match partially.

image
image
image

PR #89, opened 2024-08-18
Live reload for Gradio demo

### LINKED ISSUE #80: Update gradio demo to support streaming output

juberti:
Fixes #80 

Now, changes that you make during editing are reflected immediately in the demo. For this to work `demo` has to be a global object, so the `main` function has been removed (as recommended by gradio: https://www.gradio.app/guides/quickstart)

Also adds streaming of output tokens, using `infer_stream` and queueing.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -55,7 +55,7 @@ mds *FLAGS:
     poetry run python -m ultravox.tools.mds_tool {{FLAGS}}
 
 gradio *FLAGS:
-    poetry run python -m ultravox.tools.gradio_demo {{FLAGS}}
+    poetry run gradio ultravox/tools/gradio_demo.py {{FLAGS}}
 
 run *FLAGS:
     poetry run mcli run -f mcloud.yaml --follow {{FLAGS}}
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -190,7 +190,6 @@ class LocalInference(base.VoiceInference):
             do_sample=do_sample,
             max_new_tokens=max_new_tokens or MAX_NEW_TOKENS,
             temperature=temperature,
-            repetition_penalty=REPETITION_PENALTY,
             pad_token_id=self.tokenizer.eos_token_id,
             eos_token_id=terminators,
             streamer=streamer,
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -60,7 +60,7 @@ class LocalInference(base.VoiceInference):
         audio_token_len: int,
         response_content: str,
     ) -> List[Dict[str, str]]:
-        messages = query_messages[:]
+        messages = copy.copy(query_messages)
         if audio_token_len > 0:
             user_content = messages[-1]["content"]
             if user_content.count("<|audio|>") != 1:
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -14,8 +14,6 @@ from ultravox.model import ultravox_processing
 
 SAMPLE_RATE = 16000
 MAX_NEW_TOKENS = 1024
-# Without this penalty, the model tends to repeat itself.
-REPETITION_PENALTY = 1.1
 
 
 class LocalInference(base.VoiceInference):
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -1,6 +1,6 @@
 import copy
-import queue
 import threading
+from concurrent import futures
 from typing import Dict, List, Optional, Tuple, Union
 
 import librosa
@@ -109,14 +109,14 @@ class LocalInference(base.VoiceInference):
             self.tokenizer, skip_prompt=True, skip_special_tokens=True
         )
 
-        def thunk(q: queue.Queue):
+        def thunk(f: futures.Future):
             result = self._generate(
                 inputs, max_tokens, temperature, streamer, self.past_key_values
             )
-            q.put(result)
+            f.set_result(result)
 
-        result_queue: queue.Queue = queue.Queue()
-        thread = threading.Thread(target=thunk, args=(result_queue,))
+        future = futures.Future()
+        thread = threading.Thread(target=thunk, args=(future,))
         thread.start()
         output_text = ""
         output_token_len = 0
@@ -126,7 +126,7 @@ class LocalInference(base.VoiceInference):
                 output_token_len += 1
                 yield base.InferenceChunk(chunk)
         thread.join()
-        output = result_queue.get()
+        output = future.result()
         if self.conversation_mode:
             audio_token_len = inputs.get("audio_token_len", [0])[0]
             past_messages = self._build_past_messages(
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -105,9 +105,8 @@ class LocalInference(base.VoiceInference):
         extended_sample = self._get_sample_with_past(sample)
         inputs = self._dataproc(extended_sample)
         input_tokens = inputs["input_ids"].shape[1]
-        decode_kwargs = {"skip_special_tokens": True}
         streamer = transformers.TextIteratorStreamer(
-            self.tokenizer, skip_prompt=True, decode_kwargs=decode_kwargs
+            self.tokenizer, skip_prompt=True, skip_special_tokens=True
         )
 
         def thunk(q: queue.Queue):
--- a/ultravox/inference/infer.py
+++ b/ultravox/inference/infer.py
@@ -1,4 +1,5 @@
 import copy
+import queue
 import threading
 from typing import Dict, List, Optional, Tuple, Union
 
@@ -53,6 +54,25 @@ class LocalInference(base.VoiceInference):
         sample.add_past_messages(self.past_messages)
         return sample
 
+    def _build_past_messages(
+        self,
+        query_messages: List[Dict[str, str]],
+        audio_token_len: int,
+        response_content: str,
+    ) -> List[Dict[str, str]]:
+        messages = query_messages[:]
+        if audio_token_len > 0:
+            user_content = messages[-1]["content"]
+            if user_content.count("<|audio|>") != 1:
+                raise ValueError(
+                    f"Expected 1 audio placeholder, found {user_content.count('<|audio|>')}"
+                )
+            messages[-1]["content"] = user_content.replace(
+                "<|audio|>", self.tokenizer.eos_token * audio_token_len
+            )
+        messages.append({"role": "assistant", "content": response_content})
+        return messages
+
     def infer(
         self,
         sample: datasets.VoiceSample,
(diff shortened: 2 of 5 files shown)


juberti:
Reviewing this PR will be easier if you check "ignore whitespace" in the PR settings.

ultravox/tools/gradio_demo.py
```diff
@@ -37,116 +38,122 @@ def main():
...
+        chatbot,
+        prompt,
+        audio=audio,
+        max_new_tokens=max_new_tokens,
+        temperature=temperature,
+    )
+
+
+def gradio_reset():
+    inference.reset_conversation()
```
zqhuang211, line 102: `reset_conversation()` has been renamed to `update_conversation()` in `main`

Justfile
```diff
@@ -55,7 +55,7 @@ mds *FLAGS:
     poetry run python -m ultravox.tools.mds_tool {{FLAGS}}
 
 gradio *FLAGS:
-    poetry run python -m ultravox.tools.gradio_demo {{FLAGS}}
+    poetry run gradio ultravox/tools/gradio_demo.py {{FLAGS}}
```
zqhuang211, line 58: How do you pass arguments with this change? It looks like the inference is not running on `mps` by default on Mac, and I got the following error when specifying the `--device` argument:
```
just gradio --device mps
poetry run gradio ultravox/tools/gradio_demo.py --device mps
Usage: gradio [OPTIONS] DEMO_PATH
Try 'gradio --help' for help.
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ No such option: --device                                                                                                                                                                                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
error: Recipe `gradio` failed on line 58 with exit code 2
```

ultravox/tools/gradio_demo.py
```diff
@@ -37,116 +38,122 @@ def main():
...
+        sample,
+        max_tokens=max_new_tokens,
+        temperature=temperature,
+    )
+    chatbot += [[None, ""]]
+    for chunk in output:
+        if isinstance(chunk, infer_base.InferenceChunk):
+            chatbot[-1][1] += chunk.text
+            yield chatbot
+
```
zqhuang211, line 83: `infer_stream` doesn't support `past_message` and `past_key_values` yet, so the model isn't aware of the conversation history and treats each conversation turn as an independent request.

ultravox/tools/gradio_demo.py
```diff
@@ -37,116 +38,122 @@ def main():
...
+            )
+            temperature = gr.Slider(
+                minimum=0,
+                maximum=5.0,
+                value=args.temperature,
+                step=0.1,
+                interactive=True,
+                label="temperature",
+            )
+    prompt.submit(add_text, [chatbot, prompt], [chatbot, prompt], queue=False).then(
```
zqhuang211, line 143: should `queue=False` be removed to use the `queueing` setting of the app (enabled by `demo.queue()`?

juberti: Let me look more closely. The streaming works without this change, but maybe we should change this anyway now that we have queuing, as you suggest.

Justfile
juberti: Hmm, it seems like it should work, per https://www.gradio.app/guides/developing-faster-with-reload-mode#python-ide-reload, but I'm getting the same error as you. I've asked on the Gradio Discord to try to figure out what we can do here, but in the short term I can try to make sure `mps` is used by default on Mac.

ultravox/tools/gradio_demo.py
juberti: The docs at https://www.gradio.app/guides/creating-a-custom-chatbot-with-blocks seem to suggest that queueing should only be used for operations that will take a while, so I think this usage is OK here.

Justfile
juberti: Updated to make mps the default on Mac, as it seems to work pretty well now.

ultravox/tools/gradio_demo.py
juberti: Right, I had forgotten. Fixed.

ultravox/tools/gradio_demo.py
```diff
@@ -37,116 +38,122 @@ def main():
...
-        chatbot = chatbot + [(None, output.text)]
-        return chatbot, gr.update(value=prompt_to_return)
 
-    def process_text(chatbot, prompt, max_new_tokens, temperature):
-        return process_turn(
-            chatbot, prompt, max_new_tokens=max_new_tokens, temperature=temperature
-        )
+def add_audio(chatbot: gr.Chatbot, audio: str) -> gr.Chatbot:
+    # We want to keep the prompt (mixed audio/text instruction) as is in voice mode.
+    return chatbot + [[(audio,), None]]
```
zqhuang211, line 50: We also want to display the prompt used in mixed audio/text instruction, as it provides the full context in the chatbot.

```suggestion
def add_audio(chatbot: gr.Chatbot, audio: str, text: str) -> gr.Chatbot:
    # We want to keep the prompt (mixed audio/text instruction) as is in voice mode.
    if text:
        return chatbot + [[text, None]] + [[(audio,), None]]
    else:
        return chatbot + [[(audio,), None]]
```

ultravox/tools/gradio_demo.py
```diff
@@ -37,116 +38,122 @@ def main():
...
+                step=0.1,
+                interactive=True,
+                label="temperature",
+            )
+    prompt.submit(add_text, [chatbot, prompt], [chatbot, prompt], queue=False).then(
+        process_text,
+        [chatbot, prompt, max_new_tokens, temperature],
+        [chatbot],
+    )
+    audio.stop_recording(add_audio, [chatbot, audio], [chatbot], queue=False).then(
```
zqhuang211, line 148: Also include prompt as it needs to be added to the chatbot if not empty.

```suggestion
    audio.stop_recording(add_audio, [chatbot, audio, prompt], [chatbot], queue=False).then(
```

ultravox/inference/infer.py
```diff
@@ -68,55 +88,53 @@ def infer(
         output_tokens = output.sequences[0][input_len:]
         output_text = self.tokenizer.decode(output_tokens, skip_special_tokens=True)
```
zqhuang211, line 89: I still see `<|eot_id|>` is gradio chat window, e.g.,
```
It's nice to meet you. Is there something I can help you with or would you like to chat?<|eot_id|>
```

juberti: I see this as well, but I see that even with non-streaming inference. I think there's a separate problem with this code in `_dataproc`: @farzadab 
```
      if "<|eot_id|>" in self.tokenizer.added_tokens_encoder:
            terminators.append(self.tokenizer.convert_tokens_to_ids("<|eot_id|>"))
```

juberti: maybe we're not setting `added_tokens_encoder` anymore?

farzadab: re `<|eot_id|>` in `terminators`: I don't think having the terminator means it shouldn't be in the output.
I assumed the way we used to remove that was with `skip_special_tokens=True`. Not sure if that behaviour is working as expected now or not.

ultravox/inference/infer.py
```diff
@@ -68,55 +88,53 @@ def infer(
...
+        result_queue: queue.Queue = queue.Queue()
+        thread = threading.Thread(target=thunk, args=(result_queue,))
         thread.start()
-        output_tokens = 0
+        output_text = ""
+        output_token_len = 0
         for chunk in streamer:
             if chunk:
+                output_text += chunk
+                output_token_len += 1
```
farzadab, line 127: Not really relevant to PR, but is every token representable in text?

I have a vague memory of some binary/unicode tokens (e.g. Chinese?) could not be decoded into actual text with `tiktoken` until their "pair" or "suffix" was completed.

I could be wrong of course and even then it's a very unlikely event.

ultravox/inference/infer.py
```diff
@@ -68,55 +88,53 @@ def infer(
...
                 yield base.InferenceChunk(chunk)
-                output_tokens += 1
-        yield base.InferenceStats(input_tokens, output_tokens)
         thread.join()
+        output = result_queue.get()
+        if self.conversation_mode:
+            audio_token_len = inputs.get("audio_token_len", [0])[0]
+            past_messages = self._build_past_messages(
+                extended_sample.messages, audio_token_len, output_text
+            )
```
farzadab, line 135: Try the following to see if `<|eot_id|>` is removed.
```suggestion
            output_tokens = output.sequences[0][input_len:]
            output_text = self.tokenizer.decode(output_tokens, skip_special_tokens=True)
            past_messages = self._build_past_messages(
                extended_sample.messages, audio_token_len, output_text
            )
```

ultravox/inference/infer.py
```diff
@@ -68,55 +88,53 @@ def infer(
...
-            streamer,
-        )
-        thread = threading.Thread(target=self._generate, args=thread_args)
+        def thunk(q: queue.Queue):
+            result = self._generate(
+                inputs, max_tokens, temperature, streamer, self.past_key_values
+            )
+            q.put(result)
+
+        result_queue: queue.Queue = queue.Queue()
```
farzadab, line 119: nit: can't a `Future` do the same thing? A queue of size 1 seems a bit weird to me.

farzadab (commented):
I didn't get to try the demo yet. The changes looked good though. I'll let ZQ be the final arbiter.

ultravox/tools/gradio_demo.py
```diff
@@ -28,7 +29,7 @@ class DemoConfig:
     temperature: float = 0
 
 
-def main():
```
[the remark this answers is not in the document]
juberti: perhaps, but it seemed out of step with the patterns that Gradio uses in their docs, so I figured, why fight the system.

ultravox/inference/infer.py
juberti: yeah, that might be a bit simpler.

ultravox/inference/infer.py
juberti: yeah that works. Not sure why the Streamer args aren't being respected, looking closer.

juberti: I misunderstood how the kwargs were supposed to work here, fixed.

ultravox/inference/infer.py
juberti: Fixed in the latest rev.

ultravox/inference/infer.py
juberti: Is the concern that we shouldn't be appending to `output_text`, or that `output_token_len` will be off?

ultravox/inference/infer.py
```diff
@@ -53,6 +54,25 @@ def _get_sample_with_past(
         sample.add_past_messages(self.past_messages)
         return sample
 
+    def _build_past_messages(
+        self,
+        query_messages: List[Dict[str, str]],
+        audio_token_len: int,
+        response_content: str,
+    ) -> List[Dict[str, str]]:
+        messages = query_messages[:]
```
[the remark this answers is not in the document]
juberti: I wasn't worried about it since we already copy the input messages in `infer`, but I changed it to make a copy here just to avoid any future issues.

zqhuang211 (approved):
Streaming output looks very nice!

ultravox/inference/infer.py
farzadab: Nice!

ultravox/inference/infer.py
farzadab: Not sure what would happen in this case tbh. Since a chunk was already generated, then I guess the only possibility is that the `output_token_len` will be off, but I'm not really sure if this can even happen.
Just thought to raise this in case we come across it later on. Feel free to resolve.

Issue #83, opened 2024-08-15
Support num_epochs in config


zqhuang211:
The `num_epochs` configuration parameter, intended to activate when `max_steps=0`, is currently causing model training to break.

Need to investigate the underlying cause and implement a fix.

juberti:
I think this is due an incompatibility with `streaming=True`, where the length of the epoch is not known in advance.

liPatrick:
Fix: https://github.com/fixie-ai/ultravox/pull/90

You need to have an estimate of the dataset size ahead of time for this to work.

PR #90, opened 2024-08-20
Support num epochs in config

liPatrick:
Fixes issue: https://github.com/fixie-ai/ultravox/issues/83

The problem was that when max_samples = 0, the length of the dataset gets invoked, but we didn't have any implementation for the __len__ functions for our dataset classes.


juberti:
Add tests please 😄

ultravox/data/datasets.py
```diff
@@ -1167,3 +1176,8 @@ def __iter__(self):
             if self._num_samples is not None and i >= self._num_samples:
                 break
             yield sample
+
+    def __len__(self):
+        return (
+            self._num_samples if self._num_samples is not None else len(self._dataset)
```
zqhuang211, line 1182: We may need to add the following in `__init__`, otherwise, `__len__` is not correct if `num_samples` is larger than the size of the dataset. 

```
self._num_samples = min(num_samples, len(self._dataset))

File: ultravox/data/datasets.py [hunk 2]
```diff
@@ -1138,6 +1141,9 @@ def __iter__(self):
                 iters[iter_index] = iter(self._datasets[iter_index])
                 yield next(iters[iter_index])
 
+    def __len__(self):
+        return sum(len(ds) for ds in self._datasets)
+
```
zqhuang211, line 1146: The definition of epoch for the interleaved dataset is a bit tricky due to the weighting of datasets and the stop strategy. I guess the logic used here is likely the most accurate, especially if we consider concatenation as a special case of an interleaved dataset.

Currently, concatenation can be simulated by manually setting the weights of each dataset based on its length. Perhaps we could consider using weight=0 as the default, and if any weight is set to 0, it should trigger concatenation simulation. This could be addressed in a separate PR.

ultravox/data/datasets.py
liPatrick: I think we probably need to include the actual data length somewhere in the config, or else we have to iterate through the datset to get the length.

ultravox/data/datasets.py
liPatrick: I think this depends on how huggingface uses len(train_dataset)? I did some digging and found one place where they use the length of the dataset to determine the number of shards to batch in training. Seems like that would mean we'd need the actual* size of the dataset, including the logic in interleave, and concatenation is not ideal here for strategys like last_exauhsted because it would underestimate the length.

ultravox/data/datasets.py
juberti: This is the fundamental problem with using epochs and streaming, AFAICT HF doesn't know the dataset length up front (even if you look at the info in the dataset card, it might be wrong). For datasets that we rely on I suppose we could add a hardcoded length in our wrapper, but this will naturally be somewhat fragile.

zqhuang211: I did some searching on this topic, and it looks like there’s no way to get around this with streaming mode. Right now, I rely on the info in the dataset card/readme file to find the number of rows. Maybe we just have to live with it.

ultravox/data/datasets.py
```diff
@@ -1006,7 +1005,12 @@ def __init__(
         if self._args.shuffle:
             dataset = dataset.shuffle(seed=self._args.shuffle_seed)
 
+        self.total_samples = config.total_samples
```
zqhuang211, line 1008: should we enforce `config.total_samples` always being set?

ultravox/data/datasets.py
```diff
@@ -1044,6 +1048,12 @@ def _get_sample(self, row) -> VoiceSample:
             audio_transcript=transcript,
         )
 
+    def __len__(self):
+        try:
+            return len(self._dataset)
+        except:
+            return self.total_samples
+
```
zqhuang211, line 1056: I think it would be better to ensure consistency between `config.total_samples` and `len(self._dataset)` within `__init__`.

- If `len(self._dataset)` is functional, `config.total_samples` should equal `len(self._dataset)`; otherwise, an exception should be raised so that we can catch it and update the experiment configuration file accordingly.
- If `len(self._dataset)` is not functional, we should set `self.total_samples = config.total_samples` and introduce a mechanism to count the actual number of samples iterated in the dataset. If the actual number (computed when a `StopIteration` exception is thrown) is less than `self.total_samples`, or if it exceeds `self.total_samples` at any point during iteration, we should print a warning message to avoid wasting compute time.

ultravox/data/datasets_test.py
```diff
@@ -295,3 +305,14 @@ def test_get_messages():
         {"role": "user", "content": "B"},
         {"role": "assistant", "content": "C"},
     ]
+
+
+def test_get_dataset_len():
+    ds_1 = FakeRange(FakeIterableDataset(5), 5)
```
juberti, line 311: Can you use the real classes here? Not sure what the Fakes are needed for.

ultravox/data/datasets.py
juberti: What's the plan to make this work for datasets other than `GenericVoiceDataset`?

liPatrick:
I'm going to add a wrapper class around iterableDataset that gives an estimated length of the iterable dataset (which we instantiate it with), and whenever we iterate through the dataset, we keep a counter of the length and log when we reach the end of the iterable and the length doesnt match the instantiated length.

ultravox/data/datasets.py
liPatrick: I think in the near future, every dataset will either be able to be derived from GenericVoiceDataset, or it will inherit from it.  Also, other datasets, don't have the ability to use epochs right now, which is what implementing __len__ is primarliy for right now.

liPatrick: But either way, I think my SizediterableDataset implmenation should work for other datasets as well

ultravox/data/datasets.py
```diff
@@ -961,6 +1002,7 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
         dataset = self._load_audio_dataset(
             "fixie-ai/soda-audio", split=args.split.value
         )
+        dataset = SizedIterableDataset(dataset, 0)
```
liPatrick, line 1005: Setting the size here to 0 for now. The length doesn't matter for non GenericVoiceDataset types because they can't use epoch. Also, planning on getting rid of these classes/extending to use GenericVoiceDataset soon anyways.

ultravox/data/datasets.py
```diff
@@ -266,7 +267,32 @@ def _get_messages(
...
+
+    def __init__(self, dataset: data.Dataset, estimated_length: int) -> None:
+        self._dataset = dataset
+        self._estimated_length = estimated_length
+
+    def __iter__(self):
+        self._actual_length = 0
+        for item in self._dataset:
+            self._actual_length += 1
+            yield item
```
zqhuang211, line 283: Should there be a warning when `self._actual_length` becomes greater than `self._estimated_length`?

ultravox/data/datasets.py
```diff
@@ -266,7 +267,32 @@ def _get_messages(
...
+
+        if self._actual_length != self._estimated_length:
+            warnings.warn(
+                f"Mismatch between estimated length ({self._estimated_length}) "
+                f"and actual length ({self._actual_length}) for dataset of type {type(self._dataset)}"
+            )
+
```
zqhuang211, line 290: If we add the previous warning, we should only trigger the warning if `self._actual_length < self._estimated_length`

ultravox/data/datasets.py
```diff
@@ -275,13 +301,14 @@ class VoiceDataset(abc.ABC, data.IterableDataset):
...
-        super().__init__()
+        abc.ABC.__init__(self)
         self._args = args
         self._session: Optional[requests.Session] = None
         self._rng = np.random.default_rng(self._args.shuffle_seed)
         self._weight = 1.0  # the default weight for the dataset
 
-    def _init_dataset(self, dataset: data.Dataset) -> None:
+    def _init_dataset(self, dataset: SizedIterableDataset) -> None:
+        SizedIterableDataset.__init__(self, dataset, len(dataset))
```
zqhuang211, line 311: If dataset is already a `SizedInterableDataset`, why `__init__` is needed here?

ultravox/data/datasets.py
```diff
@@ -1001,6 +1042,10 @@ def __init__(
                 for s in config.splits
             ]
         )
+        if not config.total_samples:
+            raise ValueError("total_samples must be set in the dataset config")
+
+        dataset = SizedIterableDataset(dataset, config.total_samples)
 
         # shuffling is only supported on huggingface datasets for now, not MDS
         if self._args.shuffle:
```
zqhuang211, line 1051: @farzadab does it make sense to shuffle datasets with `streaming=True`, if so, how was it done?

ultravox/data/datasets.py
```diff
@@ -1153,17 +1207,34 @@ def __iter__(self):
         return (self._process(sample) for sample in self._dataset)
 
 
-class Range(data.IterableDataset):
+class Range(SizedIterableDataset):
     """Limits the number of samples from another dataset."""
 
     def __init__(
-        self, dataset: data.IterableDataset, num_samples: Optional[int] = None
+        self, dataset: SizedIterableDataset, num_samples: Optional[int] = None
```
zqhuang211, line 1214: What's the reason to allow `num_samples = None`? Conceptually, I thought `num_samples` should always be nonnegative and no greater than `len(self._dataset)`

farzadab: This allows using `Range`, not just when `num_samples` is provided, but always. So the options are: the full dataset or an artificially restricted dataset. So `Range(ds, num_samples=None)` is the same as `ds`.

ultravox/data/datasets.py
liPatrick: this is so we can use the len() and iter function from the sizeiterabledataset. if we dont init, we have to implement those methods ourselves in this class.

ultravox/data/datasets.py
liPatrick: Yes, good point I'll add that in.

ultravox/data/datasets.py
zqhuang211: I see there are places we don't check `num_samples` before using `Range`.

ultravox/data/datasets.py
farzadab: Yes it does. HF takes care of that and I've not looked at the implementation.

Theoretically, on the client-side you can use a concept called a "shuffle buffer" which loads a larger set of samples and then samples are take by random from that buffer instead. This approach is an approximate shuffle not a true shuffle for large datasets.

HF, however, owns both the client and the server, so they can use a smarter approach.

zqhuang211 (approved):

ultravox/data/datasets.py
```diff
@@ -1153,17 +1207,34 @@ def __iter__(self):
...
+        limit = (
+            self._num_samples if self._num_samples is not None else len(self._dataset)
+        )
+
+        for item in self._dataset:
+            if count >= limit:
                 break
-            yield sample
+            yield item
+            count += 1
```
farzadab, line 1235: I see two issues here:
1. this is breaking the original interface of Range when num_samples is not specified and I'm not sure why. And in this case any dataset whose length is not specified correctly is gonna be incorrectly subsampled (in many cases to 0 samples, i.e. deleted)

2. I don't see why the main code was changed here. You could simply do `num_samples = ...` and keep the original code.

ultravox/data/datasets.py
```diff
@@ -266,7 +267,32 @@ def _get_messages(
...
+class SizedIterableDataset(data.IterableDataset):
+    """
+    A wrapper for a Dataset that provides a length method and verifies the count.
+    """
+
+    def __init__(self, dataset: data.Dataset, estimated_length: int) -> None:
+        self._dataset = dataset
+        self._estimated_length = estimated_length
+
+    def __iter__(self):
```
farzadab, line 279: When is this function ever used?
From what I see this is supposed to be a superclass of `VoiceDataset` which already has an `__iter__` function so it gets overwritten.

ultravox/data/datasets.py
```diff
@@ -768,6 +803,7 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
             )
         if self._args.shuffle:
             ds = ds.shuffle(seed=self._args.shuffle_seed)
+        ds = SizedIterableDataset(ds, 0)
```
farzadab, line 806: What's exactly the point of using 0 here?
Also, I thought `SizedIterableDataset` was supposed to be the parent class. Why use it both as a parent class and as a wrapper.

ultravox/data/datasets.py
farzadab: It does matter. Combined with the `Range` changes this code is gonna remove any non-generic datasets.

ultravox/data/datasets.py
```diff
@@ -1001,6 +1042,10 @@ def __init__(
                 for s in config.splits
             ]
         )
+        if not config.total_samples:
+            raise ValueError("total_samples must be set in the dataset config")
```
farzadab, line 1046: If `total_samples` is required, then enforce it in the dataclass instead of raising an error here.

ultravox/data/datasets.py
```diff
@@ -1093,8 +1138,12 @@ def __init__(
             stop_strategy: Strategy for stopping iteration.
             seed: Optional seed for reproducibility.
             static: If true, the datasets are interleaved in a static order with equal weights.
+
+        Note: This class inherits from SizedIterableDataset for type compatibility,
+        but implements its functionality independently. It does not call
+        super().__init__() as it operates on multiple datasets rather than a single one,
+        and it doesn't rely on the parent's implementation.
         """
-        super().__init__()
```
farzadab, line 1097: Failing to call `__init__` is generally not a good idea and seems to be a bad smell to me.

In this particular case it seems you can simply call `len(self)` once `self._datasets` is set to get num_samples.

`self._dataset` however is never set because of this and that means this class isn't a sub-class of `SizedIterableDataset`.

farzadab (changes requested):

ultravox/data/datasets.py
farzadab: I see this class is used both as a parent class and as a wrapper. That solves this issue I guess ...

ultravox/data/datasets.py
liPatrick: Yes, good catch, I assumed the len() invariant incorrectly here.

ultravox/data/datasets.py
liPatrick: Updated, the Range code.

ultravox/data/datasets.py
liPatrick: Changed the class into an interface, and implemented the len and iter in the voicesample class.

farzadab (approved):
Looks much cleaner, thanks!

zqhuang211:
@liPatrick before merging to main, can you first try to run an experiment similar to v0.3 to make sure everything works as expected?

liPatrick:
A run of ultravox v3 using epochs: 
https://wandb.ai/fixie/ultravox/runs/bvbdxz0v/overview

PR #92, opened 2024-08-20
Working AutoProcessor.from_pretrained

farzadab:
With this PR (assuming the code is updated on HF), we can call `transformers.AutoProcessor.from_pretrained('fixie-ai/ultravox-v0_3')` to get the processor (needed by @petersalas for the VLLM PR).

### DIFF
--- a/ultravox/model/ultravox_processing.py
+++ b/ultravox/model/ultravox_processing.py
@@ -4,6 +4,8 @@ import numpy as np
 import torch
 import transformers
 
+from .ultravox_config import UltravoxConfig
+
 
 class UltravoxProcessor(transformers.ProcessorMixin):
     """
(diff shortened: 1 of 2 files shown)


ultravox/model/ultravox_processing.py
```diff
@@ -175,3 +200,6 @@ def model_input_names(self):
         tokenizer_input_names = self.tokenizer.model_input_names
         audio_processor_input_names = self.audio_processor.model_input_names
         return list(set(tokenizer_input_names + audio_processor_input_names))
+
+
+transformers.AutoProcessor.register(UltravoxConfig, UltravoxProcessor)
```
petersalas, line 205: Nice! We should presumably remove the commented out bit in the other file too :)

petersalas (approved):
Thanks!

PR #96, opened 2024-08-22
Change tokenizer padding_side to left for eval

zqhuang211:
`text_tokenizer.padding_side=right` is used during model training but needs to be changed to `left` (as expected by the `LocalInference` class to support batch inference and for consistency) for model inference at the end of training run.

### DIFF
--- a/ultravox/training/train.py
+++ b/ultravox/training/train.py
@@ -303,6 +303,8 @@ def main() -> None:
         # Merge LoRA weights for better inference performance.
         # Note: this is irreversible and changes model saving format
         model.merge_and_unload()
+        # padding_side="left" is required for (batch) inference
+        text_tokenizer.padding_side = "left"
         inference = infer.LocalInference(
             model=model,
             processor=processor,


ultravox/training/train.py
```diff
@@ -303,6 +303,8 @@ def main() -> None:
         # Merge LoRA weights for better inference performance.
         # Note: this is irreversible and changes model saving format
         model.merge_and_unload()
+        # padding_side="left" is required for (batch) inference
+        text_tokenizer.padding_side = "left"
```
farzadab, line 307: We use batch_size of 1 and no collator, so this should not be needed.

zqhuang211: `left` is enforced in the `__init__` method

https://github.com/fixie-ai/ultravox/blob/f2fbdafd5e6eecbd22b70431b3bd7cf5da60d47e/ultravox/inference/infer.py#L44

farzadab: Oh okay. Seems like the assert was wrong to put in then. I was confused too when Pat said `UltravoxInference` does the left padding but enforced `padding_left` in `LocalInference`.

We could also move the assert to the start of the `infer_batch` function.

But no strong opinion.

zqhuang211: Agreed. On the other hand, I think it is better to keep it consistent for the `LocalInference` in terms of `padding_side`.

farzadab (approved):

liPatrick (approved):

PR #97, opened 2024-08-23
Add chunking to ds_tool

liPatrick:
Adding chunking to ds_tool so it is more resilient to processing/upload failures for larger datasets. 

Dynamically chunks dataset to process and upload. Failed chunks get reprocessed into smaller chunks and recursively go through the same process. Repeated until the chunk_split_threshold is reached, then it gets uploaded as is. 

Note:
Push_to_hub automatically shards chunks into roughly 500mb. So in huggingface there will be a chunk and shard denominator, but we can just use a wildcard to set the path for our splits in the README, ie subset_name/split_name/**


ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -210,6 +211,9 @@ class DatasetToolArgs:
     private: bool = simple_parsing.field(default=False)
     token: Optional[str] = None
 
+    # Chunk processing parameters
+    max_samples_per_chunk: int = simple_parsing.field(default=100000)
```
liPatrick, line 215: I'm choosing max_samples isntead of max_size because our processing failures are usually correlated with sample # not sample size.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -210,6 +211,12 @@ class DatasetToolArgs:
     private: bool = simple_parsing.field(default=False)
     token: Optional[str] = None
 
+    # Chunk processing parameters
+    max_dynamic_chunks: int = simple_parsing.field(default=10)
+    min_samples_per_chunk: int = simple_parsing.field(
+        default=10000
```
liPatrick, line 217: I'm choosing sample number instead of sample size because our processing failures are usually correlated with # of sample not sample size.

ultravox/data/text_proc.py
```diff
@@ -31,15 +40,17 @@ def format_asr_text(text: str) -> str:
     remaining_words = []
     for word in text.split():
         if word in GIGASPEECH_GARBAGE_UTTERANCE_TAGS:
-            return ""
+            raise GarbageUtteranceError(f"Garbage utterance tag found: {word}")
```
zqhuang211, line 43: This will remove the utterance all together. I think it is better than "", but we may want to revisit this and include these utterances as they may help improve the robustness of the model.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -143,20 +144,26 @@ def _map_sample(self, sample, exclude_fields):
...
+            if isinstance(e, text_proc.GarbageUtteranceError):
                 print("Formatted text is empty. Setting output to None.")
                 sample[self.new_column_name] = None
                 self.template_failures += 1
-            else:
+                return sample
+            elif isinstance(e, text_proc.EmptyTranscriptError):
+                print("Empty transcript after processing. Setting output to None.")
+                sample[self.new_column_name] = None
+                self.template_failures += 1
```
zqhuang211, line 156: The name `template_failures` is a bit confusing (considering the actual `jinja2.TemplateError` in line 158).

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -189,7 +196,7 @@ def _map_sample(self, sample, exclude_fields):
 class DatasetToolArgs:
     # HF source dataset parameters
     dataset_name: str = simple_parsing.field(alias="-d")
-    dataset_subset: Optional[str] = simple_parsing.field(default=None, alias="-S")
+    dataset_subset: Optional[str] = simple_parsing.field(default="default", alias="-S")
```
zqhuang211, line 199: What happens if the dataset only has one subset/config whose name is not `default`, when calling `datasets.load_dataset` without specifying the subset/config?

ultravox/data/text_proc.py
farzadab: I don't see this error being handled in `datasets.py`, that means this is going to break training of any dataset that uses this function. Same for the other error.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+            split_name, ds_split, 0, len(ds_split)
+        )
+
+        while len(failed_chunk_ranges) > 0:
+            new_failed_ranges = []
+            for start, end in failed_chunk_ranges:
+                print(f"Retrying failed chunk range [{start}, {end})")
+                new_failed_ranges.extend(
+                    self._dynamic_chunk(split_name, ds_split, start, end)
+                )
```
farzadab, line 251: Seems like you can have a dataset where only a portion of the samples are uploaded, is that true?

I need to read more of the code, but that seems wrong.
1. Does the order of samples change?
3. Is having a half-finished dataset okay? What was the trade-off, e.g. why waiting to process then upload (with retry) doesn't work?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+            failed_chunk_ranges = new_failed_ranges
+        print(f"Could not upload chunks: {self.chunks_not_uploaded}")
+        print(
+            f"Finished processing and uploading all chunks for split {split_name}. Total samples processed: {self.total_samples_processed}"
+        )
+
+    def _dynamic_chunk(
```
farzadab, line 258: Function name doesn't suggest that uploading is also happening.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+        split_name: str,
+        ds_split: datasets.Dataset,
+        start_index: int,
+        end_index: int,
+    ):
+        original_chunk_size = end_index - start_index
+        if original_chunk_size < self.args.chunk_split_threshold:
+            total_chunks = 1
+            chunk_size = original_chunk_size
+            print(
```
farzadab, line 269: nit: I feel like we can start to use `logging` instead of print. Not necessary for this PR though.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+        else:
+            total_chunks = self.args.max_chunk_split
+            chunk_size = math.ceil(original_chunk_size / total_chunks)
+        failed_chunk_ranges = []
+        print(
+            f"Processing and uploading {total_chunks} chunks for range [{start_index}, {end_index}) with chunk size {chunk_size}"
+        )
+        for i in range(start_index, end_index, chunk_size):
+            chunk_start = i
+            chunk_end = min(i + chunk_size, start_index + original_chunk_size)
```
farzadab, line 281: nit:
```suggestion
            chunk_end = min(i + chunk_size, end_index)
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+            )
+        else:
+            total_chunks = self.args.max_chunk_split
+            chunk_size = math.ceil(original_chunk_size / total_chunks)
+        failed_chunk_ranges = []
+        print(
+            f"Processing and uploading {total_chunks} chunks for range [{start_index}, {end_index}) with chunk size {chunk_size}"
+        )
+        for i in range(start_index, end_index, chunk_size):
+            chunk_start = i
```
farzadab, line 280: nit:
```suggestion
        for chunk_start in range(start_index, end_index, chunk_size):
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+            chunk_end = min(i + chunk_size, start_index + original_chunk_size)
+
+            ds_chunk = ds_split.select(range(chunk_start, chunk_end))
```
farzadab, line 283: This might not be the right thing here due to the recursion, but just asking if you've seen [`dataset.shard`](https://huggingface.co/docs/datasets/en/process#shard).
Example: `dataset.shard(num_shards=4, index=0)`

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+        print(f"Uploading chunk to hub: {data_dir}")
+        hub_args: Dict[str, Any] = {
+            "config_name": self.args.upload_subset,
+            "token": self.args.token or os.environ.get("HF_TOKEN"),
+            "private": self.args.private,
+            "data_dir": data_dir,
+            "num_shards": self.args.num_shards,
+            "split": split_name,
+        }
+        ds_chunk_processed.push_to_hub(self.args.upload_name, **hub_args)
```
farzadab, line 346: Are you absolutely sure this doesn't override the split? I'm fairly confident that it would. Even if it was "appending" to it, `num_shards` would be totally wrong I think.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+                        f"Skipping chunk {ds_chunk_name} as it has already been processed and uploaded."
+                    )
+                    ds_chunk_processed = datasets.Dataset.from_parquet(
+                        ds_chunk_cache_path
+                    )
+                else:
+                    print(f"Processing chunk {ds_chunk_name}")
+                    ds_chunk_processed = self._process(ds_chunk)
+                    self._upload(ds_chunk_processed, ds_chunk_hub_path, split_name)
+                    ds_chunk_processed.to_parquet(ds_chunk_cache_path)
```
farzadab, line 307: Won't you loose the chunk if upload fails? I suggest switching the order:
```suggestion
                    ds_chunk_processed.to_parquet(ds_chunk_cache_path)
                    self._upload(ds_chunk_processed, ds_chunk_hub_path, split_name)
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+                self.total_samples_processed[split_name] = (
+                    self.total_samples_processed.get(split_name, 0)
+                    + len(ds_chunk_processed)
+                )
```
farzadab, line 311: Total nit and mostly just syntactic sugar, but I think what you want is `defaultdict(int)`:
```python
>>> collections.defaultdict(int)
defaultdict(<class 'int'>, {})
>>> d = collections.defaultdict(int)
>>> d[10] += 2
>>> d
defaultdict(<class 'int'>, {10: 2})
```

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +228,124 @@ def __post_init__(self):
...
+            except Exception as e:
+                print(f"Failed to upload chunk {ds_chunk_name}: {e}. Retrying later.")
+                if total_chunks == 1:
+                    print(
+                        f"Finished processing and uploading 0/1 chunks for range [{start_index}, {end_index})"
+                    )
+                    self.chunks_not_uploaded.append((start_index, end_index))
+                    return []
+                failed_chunk_ranges.append((chunk_start, chunk_end))
+        successful_chunks = self.args.max_chunk_split - len(failed_chunk_ranges)
```
farzadab, line 322: I'm confused about the logic here.

1. I don't get what `self.chunks_not_uploaded` does. It only seems to be used for logging. So if you only have 1 chunk and you fail, then you're done and you won't retry?

2. IIRC the logic you said you wanted to implement was to recursively use smaller and smaller chunk sizes until you can do it in one go. I don't think I see that recursion here. The only chunking I see is via `chunk_split_threshold` which is static.

3. Having special logic for `total_chunks == 1` vs the other chunk sizes is also a bad smell, unless there's a special reason I didn't understand.

farzadab (commented):
Loved the `DatasetChunkProcessor` abstraction. It cleanly separates the concerns without increasing complexity in other areas of the code.

I left too many nit comments, feel free to ignore most, but 2 comments are very important: 1) I feel like `_upload` would rewrite the whole set which is not desired and 2) I didn't fully see the "recursive" logic that I thought you mentioned in TGIF. I expected a different approach there.

ultravox/data/text_proc.py
farzadab: I think dropping the samples in `dataset.py` is good enough for now.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: So because I'm uploading it via data_dir, the only thing its overriding is the README (which I should leave a note that we need to update at the end of this whole thing to reflect the correct paths and split metadata).

ultravox/tools/ds_tool/ds_tool.py
liPatrick: the idea is if the upload fails, we then want to break it up into smaller chunks to upload. We actually don't really need to save the chunk here, because all we're using the saved chunk for is to check if that chunk has already been uploaded to hf. I had some other ideas before about retrying saved chunks, but figured splitting into smaller chunks would be simpler, and also handles more edge cases.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: So i guess its not strictely resursive, but the behavior is more or less recursive. The idea of _dynamic_chunk is, we split up the input chunk into n chunks. We try processing each one. if the ith chunk fails to upload, we add it to the failed chunks array. At the end we return the array The higher order function process_and_upload_split basically handles retrying the failed chunks. The base case for _dynamic_chunk doesnt split input chunk into n chunks if chunk_split_threshold is not met. This prevents us from chunking all the way down to 1 sample (if we don't want to).

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -143,20 +140,24 @@ def _map_sample(self, sample, exclude_fields):
             rendered = jinja2.Template(
                 self.template, undefined=jinja2.StrictUndefined
             ).render(**filtered_sample, json_dump=json.dumps, text_proc=text_proc)
-
-        except jinja2.TemplateError as e:
-            error_message = str(e)
-            if "The formatted text is empty." in error_message:
+        except Exception as e:
+            if isinstance(e, text_proc.GarbageUtteranceError):
```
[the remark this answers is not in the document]
liPatrick: Good point! I just realized this too, will make the change

ultravox/tools/ds_tool/ds_tool.py
liPatrick: 1. The order of samples will change if there's a failure. But it doesn't seem important to me to keep the order?
2. It won't be half-finished, will most likely be 99% finished, with the 1% unfinished being the problematic part of the dataset (ie, ill formatted data, random Null values that the filter didn't catch).

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Updated to _split_chunk_and_upload(). Maybe that's more fitting

ultravox/tools/ds_tool/ds_tool.py
liPatrick: I havn't seen this, this would be useful

ultravox/tools/ds_tool/ds_tool.py
liPatrick: num_shards would be correct for each chunk of data. but total num shards would be num_chunks*num_shards

farzadab: Oh if you have a way of updating README then it's fine.
The num_shards thing is a bit confusing tbh, but that's fine too. Just make sure to clearly comment this where the parameter is defined.

liPatrick: Yeah, I will add in a comment to update the README

ultravox/tools/ds_tool/ds_tool.py
farzadab: Hmm, that seems wasteful if you just happen to get a networking issue. But yeah I guess you mentioned sometimes these issues are persistent for some reason, is that true?

Then I suggest two things:
1. add a comment to explain why this decision was made and why switching the order will break the code. I'd make sure that the assumption "saved chunks indicate what's been uploaded to HF" is quite prominent.

2. (not required for this PR) try to add a retry for `_upload` to handle transient network failures

liPatrick: So I'm not actually loading the cached parquet anywhere. I just use it to check if the path exists and if it does then we skip processing AND uploading. So its important that it comes before the upload, or else we potentially end up in a case where the chunk is processed and cached, but not uploaded. The next time we run the program, the program skips over the process and upload step if it finds the chunk in the cache. Reason why I tie the process and upload step together is because I've noticed that reprocessing can sometimes help with uploading errors. Maybe its best to not store the whole parquet here, rather just a note in the cache that the chunk has already been uploaded to hf, but then again having the whole parquet maybe useful if something unexpected happens.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -236,47 +359,18 @@ def main(args: DatasetToolArgs):
     if len(data_dict) > 1 and args.upload_split:
         raise ValueError("Cannot upload multiple splits to a single split")
 
-    for split, ds_split in data_dict.items():
+    ds_chunk_proc = DatasetChunkProcessor(args)
+
+    for split_name, ds_split in data_dict.items():
         print(
-            f"Processing dataset: {ds_name}, subset {args.dataset_subset}, split {args.dataset_split}, containing {len(ds_split)} samples"
+            f"Processing dataset: {args.dataset_name}, subset {args.dataset_subset}, split {split_name}, containing {len(ds_split)} samples"
```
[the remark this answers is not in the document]
liPatrick: Some vestigal changes from debugging, didnt mean to do that.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -143,20 +142,17 @@ def _map_sample(self, sample, exclude_fields):
             rendered = jinja2.Template(
                 self.template, undefined=jinja2.StrictUndefined
             ).render(**filtered_sample, json_dump=json.dumps, text_proc=text_proc)
-
+        except text_proc.FormatASRError as e:
+            print(f"Format ASR Error {e}")
+            sample[self.new_column_name] = ""
```
liPatrick, line 147: Setting this to "" instead of None because of hf dataset type errors associated with setting the value to None.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -236,47 +360,24 @@ def main(args: DatasetToolArgs):
...
-
-        # If the push fails or upload_name is not specified, save the data locally.
-        for split in data_dict.keys():
-            output_name = f"{split}-00000-of-00001.parquet"
-            data_dict[split].to_parquet(output_name)
-            print(f"Saved to {output_name}")
-            print(f"Sample {0} of {split}: {data_dict[split][0]}")
+    # Note: After running this script, you need to manually update the README.md file with the new dataset information.
+    # 1. Change the configs section path to point to upload_subset/split_name/**
+    # 2. Update the dataset split num_examples field to the total number of samples processed. This can be found in the logs as the final output.
```
farzadab, line 380: Oh wait, I didn't realize this has to be done manually.
Isn't there a better way? That's a huge breaking change! That means this tool is totally incomplete on its own.

ultravox/tools/ds_tool/ds_tool.py
farzadab: Hmm, so now it's not allowed to set the column to "" which is an artificial constraint, but I don't see a hugely better solution. Using a pre-specified string might be better, but not ideal.

Please mention that "" is not allowed somewhere though.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Will see if there's a better way

ultravox/data/datasets.py
```diff
@@ -796,8 +802,11 @@ def __init__(self, args: VoiceDatasetArgs) -> None:
         )
         self._init_dataset(dataset)
 
-    def _get_sample(self, row) -> VoiceSample:
-        return self._get_transcribe_sample(row, tproc=text_proc.format_asr_text)
+    def _get_sample(self, row) -> Optional[VoiceSample]:
+        try:
+            return self._get_transcribe_sample(row, tproc=text_proc.format_asr_text)
+        except text_proc.FormatASRError:
+            return None
```
zqhuang211, line 809: It looks like all the `try..except` statements can be moved to `_get_transcribe_sample`?

ultravox/tools/ds_tool/ds_tool.py
zqhuang211: All samples with `self.new_column_name = None` will be filtered, right? If yes, maybe add a comment here to make this point explicit?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +226,127 @@ def __post_init__(self):
             self.upload_split = self.dataset_split
 
 
+class DatasetChunkProcessor:
+    args: DatasetToolArgs
+    cache_dir: str = ".cache/ds_tool/processed_datasets"
+    chunks_not_uploaded: List[Tuple[int, int]] = []
+    total_samples_processed: Dict[str, int] = defaultdict(int)
+
```
zqhuang211, line 234: The logic and consistency in the naming used here are unclear.

If `split_name` is always required, is `total_samples_processed` a dictionary with only a single key?

What's the difference between `not_uploaded` and `processed`?  What's the difference between `total_samples_processed` vs a simpler name `samples_processed`?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +226,127 @@ def __post_init__(self):
...
+        self.args = args
+
+    def process_and_upload_split_rescursive(
+        self,
+        split_name: str,
+        ds_split: datasets.Dataset,
+        start_index: int,
+        end_index: int,
+    ):
+        original_chunk_size = end_index - start_index
```
zqhuang211, line 245: `original_` in `original_chunk_size` implies you have a new `chunk_size`, but you are not really changing `chunk_size` here. You are simply splitting a dataset into chunks and do it once.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +226,127 @@ def __post_init__(self):
...
+        if original_chunk_size < self.args.chunk_split_threshold:
+            total_chunks = 1
+            chunk_size = original_chunk_size
+            print(
+                f"Chunk range [{start_index}, {end_index}) is too small to split further. Processing and uploading as a single chunk."
+            )
+        else:
+            total_chunks = self.args.max_chunk_split
+            chunk_size = math.ceil(original_chunk_size / total_chunks)
```
zqhuang211, line 254: The logic used here is also not consistent. If you have a variable named `chunk_split_threshold`, it appears to imply that all chunks should be smaller than this number. However that is not the case. The effective `chunk_size` computed in the `else` statement can actually be greater than this number.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +226,127 @@ def __post_init__(self):
...
+        successful_chunks = self.args.max_chunk_split - len(failed_chunk_ranges)
+        print(
+            f"Finished processing and uploading {successful_chunks}/{self.args.max_chunk_split} chunks for range [{start_index}, {end_index})"
+        )
+        while len(failed_chunk_ranges) > 0:
+            new_failed_ranges = []
+            for start, end in failed_chunk_ranges:
+                print(f"Retrying failed chunk range [{start}, {end})")
+                new_failed_ranges.extend(
+                    self.process_and_upload_split_rescursive(
```
zqhuang211, line 318: I understand the logic of this code, but it is overly complicated. A alternative but simpler logic is.

1. split the dataset into chunks, where each chunk has `chunk_size` (which is a parameter) or fewer (for the last chunk) samples 
2. process the chunks one at a time, retry if necessary.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -223,6 +226,127 @@ def __post_init__(self):
...
+                        split_name, ds_split, start, end
+                    )
+                )
+            failed_chunk_ranges = new_failed_ranges
```
zqhuang211, line 322: the naming of the variables is not consistent.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Believe we resolved this over call.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: total_samples_processed can be removed now that we have huggingface README integraiton.

liPatrick: not_uploaded are chunks that failed to be processed, and could not be split further.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Think we resolved this over call as well.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -146,8 +147,7 @@ def _map_sample(self, sample, exclude_fields):
             ).render(**filtered_sample, json_dump=json.dumps, text_proc=text_proc)
         except text_proc.FormatASRError as e:
             print(f"Format ASR Error {e}. Filtering out sample.")
-            # Setting this to "" instead of None because of hf dataset type errors associated with setting the value to None.
-            sample[self.new_column_name] = ""
+            sample[self.new_column_name] = None
```
farzadab, line 150: Oh the `None` issue got fixed?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -128,7 +128,8 @@ def map_split(
 
         # Filter out samples where new_column_name is None
         return ds_mapped.filter(
-            lambda sample: sample[self.new_column_name] != "",
+            lambda sample: sample[self.new_column_name] is not None
+            and sample["audio"] is not None,
```
farzadab, line 132: Why the `sample["audio"] is not None`?
Specially since this is part of the `TextGenerationTask`.

ultravox/tools/ds_tool/chunked_dataset.py
```diff
@@ -0,0 +1,395 @@
...
+from datasets.utils.py_utils import asdict
+from datasets.utils.py_utils import glob_pattern_to_regex
+from datasets.utils.py_utils import string_to_dict
+from huggingface_hub import CommitInfo
+from huggingface_hub import CommitOperationAdd
+from huggingface_hub import CommitOperationDelete
+from huggingface_hub import DatasetCard
+from huggingface_hub import DatasetCardData
+from huggingface_hub import HfApi
+from huggingface_hub.hf_api import RepoFile
```
farzadab, line 30: nit: the import format doesn't match ours. For example, please use `import datasets` instead of `from datasets import Dataset`.

ultravox/tools/ds_tool/chunked_dataset.py
```diff
@@ -0,0 +1,395 @@
...
+
+logger = logging.get_logger(__name__)
+PUSH_TO_HUB_WITHOUT_METADATA_CONFIGS_SPLIT_PATTERN_SHARDED = (
+    "data/{split}-[0-9][0-9][0-9][0-9][0-9]-of-[0-9][0-9][0-9][0-9][0-9]*.parquet"
+)
+
+
+class ChunkedDataset(Dataset):
```
farzadab, line 38: I'm guessing most of this code was copy&pasted from `datasets` library?

I'm sure you tried to use other methods and couldn't (since maintaining this could become a pain when their new version comes out), yes?

I suggest you annotate the places where you changed the code so it's easier to reproduce later on.

farzadab (approved):
Thanks Pat.
Left some comments, but approved to unblock you.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Yes :)

ultravox/tools/ds_tool/ds_tool.py
liPatrick: I suspect training wont work with datasets that have empty audio values

farzadab: `ds_tool` was meant to be used for different tasks and even multiple times. You can first generate some stuff, then add audio for those fields or others. Even if you do have an audio, it might be in a different column.

liPatrick: Hm, then I suspect we might need to add another method that does some cleaning of the data. I can do that in a separate pr.

PR #102, opened 2024-08-28
Add AutoProcessor registration

petersalas:
UV 0.4 doesn't currently register the processor.


farzadab (approved):
Oh good catch.
That's so confusing. You'd think `AutoProcessor.register` is enough, but I guess it's not.

farzadab:
@petersalas will you update the two files (`ultravox_processin.py` and `config.json`) or should I?

petersalas:
Oh I see -- do you have the other changes we need for `from_pretrained`?

farzadab:
Done.

PR #103, opened 2024-09-04
Reverting OAI eval changes

farzadab:
As discussed with @zqhuang211, reverting to older version of eval.

W&B run to make sure things still work: https://wandb.ai/fixie/ultravox/runs/lenxya0h


ultravox/data/dataset_config.py
```diff
@@ -11,7 +11,7 @@ class DataDictConfig(BaseModel):
     name: Optional[str] = None
     splits: List[str] = dataclasses.field(default_factory=list)
     num_samples: Optional[int] = None
-    total_samples: int
+    total_samples: int = 1
```
farzadab, line 14: @liPatrick for now I just made this change to allow the default config to run since all training was broken before AFAIK. Feel free to suggest a different method or modify in follow up PRs.

liPatrick: I put up a PR to address this here: https://github.com/fixie-ai/ultravox/pull/104

farzadab: Perfect, thanks. I'll wait for that to merge then.

PR #104, opened 2024-09-04
Assert dataset length when using epochs

liPatrick:
Right now, max steps and epochs will crash with non generic datasets because their size has been set to 0 by default. As a workaround, the default size is set to 1, and an assert is added to check that real sizes are used when training with epoch.


farzadab (approved):
LGTM

PR #105, opened 2024-09-10
Replacing weight with multiplier

liPatrick:
To make running in epoch mode easier, we redefine weights to effectively be 
1. the number of times that dataset will be copied when weight > 1
2. fraction of the dataset to be used if the weight < 1. 
3. By default, the weight is 1 for all datasets. 

In the interleave class, I redefine the probability of select the next sample using the dataset length and weight to reflect the above (only in epoch mode)

### DIFF
--- a/ultravox/data/dataset_config.py
+++ b/ultravox/data/dataset_config.py
@@ -12,11 +12,9 @@ class DataDictConfig(BaseModel):
     splits: List[str] = dataclasses.field(default_factory=list)
     num_samples: Optional[int] = None
     total_samples: int = 1
-    # epochs mode:
-    #  Weight is the number of copies of the dataset
-    # max steps mode:
-    #  Weight of the dataset is used to calculate the proporition of the total samples that comes from this dataset
     weight: float = 1.0
+    # This is used over weight when epoch mode is set
+    dataset_multiplier: float = 1.0
     streaming: bool = True
     user_template: str = "<|audio|>"
     assistant_template: str = "{{text}}"
--- a/ultravox/data/dataset_config.py
+++ b/ultravox/data/dataset_config.py
@@ -1,23 +1,38 @@
 import dataclasses
+from enum import Enum
 from typing import List, Optional
 
 from pydantic import BaseModel
 
 
+class StopStrategy(str, Enum):
+    FIRST_EXHAUSTED = "first_exhausted"
+    LAST_EXHAUSTED = "last_exhausted"
+    NEVER_STOP = "never_stop"
+
+
 class DataDictConfig(BaseModel):
-    # Path to the dataset, or huggingface dataset id
-    path: str
-    # Name of the dataset, or huggingface dataset config/subset
+    path: str  # Name of the dataset, or huggingface dataset config/subset
     name: Optional[str] = None
     splits: List[str] = dataclasses.field(default_factory=list)
     num_samples: Optional[int] = None
     total_samples: int = 1
-    multiplier: float = 1.0
     streaming: bool = True
     user_template: str = "<|audio|>"
     assistant_template: str = "{{text}}"
     transcript_template: str = "{{text}}"
 
-    def __post_init__(self):
+    def post_init(self):
         if not self.splits:
             raise ValueError("At least one split must be provided")
+
+
+class DatasetMultiplier(BaseModel):
+    dataset: DataDictConfig
+    multiplier: float
+
+
+class InterleaveDataConfig(BaseModel):
+    # In InterleaveDataset, when to stop interleave: choose from last_exhausted (default), first_exhausted, or never_stop
+    stop_strategy: StopStrategy = StopStrategy.LAST_EXHAUSTED
+    datasets_with_multiplier: List[DatasetMultiplier]
--- a/ultravox/data/dataset_config.py
+++ b/ultravox/data/dataset_config.py
@@ -12,6 +12,10 @@ class DataDictConfig(BaseModel):
     splits: List[str] = dataclasses.field(default_factory=list)
     num_samples: Optional[int] = None
     total_samples: int = 1
+    # epochs mode:
+    #  Weight is the number of copies of the dataset
+    # max steps mode:
+    #  Weight of the dataset is used to calculate the proporition of the total samples that comes from this dataset
     weight: float = 1.0
     streaming: bool = True
     user_template: str = "<|audio|>"
--- a/ultravox/data/dataset_config.py
+++ b/ultravox/data/dataset_config.py
@@ -1,38 +1,23 @@
 import dataclasses
-from enum import Enum
 from typing import List, Optional
 
 from pydantic import BaseModel
 
 
-class StopStrategy(str, Enum):
-    FIRST_EXHAUSTED = "first_exhausted"
-    LAST_EXHAUSTED = "last_exhausted"
-    NEVER_STOP = "never_stop"
-
-
 class DataDictConfig(BaseModel):
-    path: str  # Name of the dataset, or huggingface dataset config/subset
+    # Path to the dataset, or huggingface dataset id
+    path: str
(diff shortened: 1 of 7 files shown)


liPatrick:
TODO: verify that we can replicate the equivalent max_steps implementation using num_epochs.

liPatrick:
> TODO: verify that we can replicate the equivalent max_steps implementation using num_epochs.

max steps: 
set to 3000
https://wandb.ai/fixie/ultravox/runs/i5hdlueh/logs

epochs: 
set to 1, with 3000 total samples (expected 16 steps because 3000 / (8 * 24)) batch size of 24, 8 gpus. 
https://wandb.ai/fixie/ultravox/runs/widpk5aw/logs

ultravox/training/configs/meta_config.yaml
```diff
@@ -1,7 +1,7 @@
 text_model: "meta-llama/Meta-Llama-3-8B-Instruct"
 audio_model: "facebook/wav2vec2-base-960h"
 
-data_sets: ["gigaspeech"]
```
liPatrick, line 4: epoch training does not work with non generic datasets. Removing this so that we can specify all datasets from release_config

ultravox/data/datasets.py
```diff
@@ -1129,6 +1130,7 @@ def __init__(
         stop_strategy: StopStrategy = StopStrategy.LAST_EXHAUSTED,
         seed: Optional[int] = 42,
         static: bool = False,
+        using_epochs: bool = False,
```
farzadab, line 1133: The name doesn't make much sense.
Maybe you want something like `use_proportional_weights` or something similar?

ultravox/data/datasets.py
```diff
@@ -1180,9 +1186,9 @@ def __iter__(self):
                 iters[iter_index] = iter(self._datasets[iter_index])
                 yield next(iters[iter_index])
 
-    def __len__(self):
+    def __len__(self) -> int:
         # TODO: Implement the length method for different stop strategies
-        return sum(len(ds) for ds in self._datasets)
+        return sum(int(getattr(ds, "weight", 1) * len(ds)) for ds in self._datasets)
```
farzadab, line 1191: Shouldn't `len` be conditional here?
IMO the concept of len doesn't make much sense when `self._using_epochs` is False.

ultravox/training/configs/meta_config.yaml
farzadab: I see `max_steps` is defined and non-zero, so we're not in epoch mode.

Setting datasets to be empty means the code won't work unless further configs are specified.
Even though most training uses `release_config`, its presence is not a must, so we cannot rely on it.

ultravox/training/train.py
```diff
@@ -48,16 +48,20 @@ def prepare_dataset(
     stop_strategy: datasets.StopStrategy,
     num_samples: Optional[int] = None,
     include_alt_fields: bool = False,  # whether to generate tensors for text-only input (e.g., used for KD training)
+    is_val_set: bool = False,
```
farzadab, line 51: The name and the function of this variable don't quite match and are confusing to me.

Just to tell you what I mean, I think a name like `enforce_ds_len_in_epoch_mode` would better describe what this variable wants to achieve (though it's too long). Also, the default of False would work better in that case since you only need to enable it when it's needed instead of opting out.

ultravox/data/datasets.py
```diff
@@ -1142,8 +1144,12 @@ def __init__(
         self._static = static
 
         self._stop_strategy = stop_strategy
+        self._using_epochs = using_epochs
+        if not self._using_epochs:
+            weights = [int(getattr(ds, "weight", 1) * len(ds)) for ds in datasets]
```
farzadab, line 1149: At this point, the term "weight" is incorrect. It seems like what you want is `data_multiplier`, not just a general notion of weight.

Another reason why, is that if you divide all "weights" by a constant, nobody expects anything to change, but in this case it does.

ultravox/data/dataset_config.py
```diff
@@ -12,6 +12,10 @@ class DataDictConfig(BaseModel):
     splits: List[str] = dataclasses.field(default_factory=list)
     num_samples: Optional[int] = None
     total_samples: int = 1
+    # epochs mode:
+    #  Weight is the number of copies of the dataset
+    # max steps mode:
+    #  Weight of the dataset is used to calculate the proporition of the total samples that comes from this dataset
```
farzadab, line 18: This distinction is very confusing and unexpected to any outsider. Using epochs or max_steps should not automatically change the behaviour other than changing the number of steps performed.

liPatrick: Added dataset_multiplier

farzadab: Thanks Pat.

Though I believe the same thing is still happening at the end of the day: the distribution of data changes when you use epoch mode or not. To add to that, now there are a set of parameters that in one scenario don't do anything but they do something in the other.

Just to make it concrete, the simplest way I see to improve on that is:
1. Have a first-class CLI argument called something like `use_proportional_dataset_multiplier` instead. I don't think this has to be tied to whether `max_steps` is 0 or not. The source of confusion is this entanglement IMO, not the name weight vs data_multiplier.
2. I believe we can remove `weight` in favour of `dataset_multiplier`. Having both and then non-transparently switching between the two is going to be a source of confusion. The reason why I think you can get away with this is that I think `dataset_multiplier` is ambiguous enough that people will not assume they know what you mean. When they do and/or when they search for `dataset_multiplier`, they will see that a `use_proportional_dataset_multiplier` arguments exists.

The naming and everything I proposed are getting a bit too long, so feel free to adjust as you see fit.

zqhuang211: I agree with Farzad that this distinction is confusing. I think `weight` should just mean `dataset_multiplier` regardless of whether we `num_epochs` or `max_steps` to determine the number of updates. `weight` had a different meaning when it was first introduced because at that time we don't have a way to measure the length of a dataset.

liPatrick: Ok, so I'll remove weight in favor of the dataset_multiplier, but that means we're removing the existing "proportional weight" functionality and if we're all fine with that I'll continue with it.

liPatrick: And I'll call it dataset_multiplier re farzad's point that 
"is that if you divide all "weights" by a constant, nobody expects anything to change, but in this case it does."

liPatrick:
With this change, we can no longer use non generic datasets (with or without epoch), because the multiplier requires a length associated with the dataset, which only generic datasets allow you to configure. I'll keep this on the sideline until https://github.com/fixie-ai/ultravox/pull/111 is merged in.

ultravox/data/datasets.py
```diff
@@ -1142,10 +1143,12 @@ def __init__(
         self._static = static
 
         self._stop_strategy = stop_strategy
+        relative_frequencies = [
```
juberti, line 1146: I think life might be easier if we didn't try to read the `multiplier` value out of each dataset class, and instead read it from a config passed to the interleave class, similar to how this is done with `probabilities` in HF's `interleave_datasets`: https://huggingface.co/docs/datasets/en/package_reference/main_classes#datasets.interleave_datasets

juberti: Then we wouldn't have to worry about the interaction with `len`, etc

liPatrick: Then we'd need to calculate the probabilities by hand in the config? I think we'd need len either way because that's how hf determines the # of steps to take in epoch mode.

juberti: Well, the weights are being set by hand already, right? It just seems a bit strange to have to read the weight/multiplier property from the dataset class, especially when the dataset class doesn't use it internally.

farzadab:
For a second there I was like, did Justin just create a PR on Ultravox public? 😄

PR #106, opened 2024-09-10
[70B-Part1] Prefetch weights separately

farzadab:
This PR allows for downloading the weights of the model first, before starting the DDP processes. The previous approach tends to time out when using 70B model.

### DIFF
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -10,7 +10,10 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run python -m ultravox.training.helpers.prefetch_weights $TRAIN_ARGS && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
+  cd ultravox &&
+  poetry install --no-dev &&
+  poetry run python -m ultravox.training.helpers.prefetch_weights $TRAIN_ARGS &&
+  poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   UV_BRANCH: main
--- a/mcloud.yaml
+++ b/mcloud.yaml
@@ -10,7 +10,7 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
+  cd ultravox && poetry install --no-dev && poetry run python -m ultravox.training.helpers.prefetch_weights $TRAIN_ARGS && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 env_variables:
   MLFLOW_TRACKING_URI: databricks
   UV_BRANCH: main
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -23,15 +23,20 @@ def main(override_sys_args: Optional[List[str]] = None):
             huggingface_hub.snapshot_download(
                 repo_id=model_id, allow_patterns=ALLOW_PATTERNS
             )
-            # A backstop to make sure the model is fully downloaded even if ALLOW_PATTERNS is not enough
-            # Using `device_map="meta"` to avoid loading the weights into memory or device
-            transformers.AutoModel.from_pretrained(model_id, device_map="meta")
         except huggingface_hub.utils.GatedRepoError as e:
             raise e
         except huggingface_hub.utils.RepositoryNotFoundError as e:
+            # We assume that the model is local if it's not found on HF Hub.
+            # The `.from_pretrained` call will verify the local case.
             print(
                 f"Model {args.text_model} not found on HF Hub. Skipping download. Error: {e}"
             )
+
+        # A backstop to make sure the model is fully downloaded. Scenarios to consider:
+        # - ALLOW_PATTERNS is not enough to download all files needed
+        # - The model is local
+        # Using `device_map="meta"` to avoid loading the weights into memory or device
+        transformers.AutoModel.from_pretrained(model_id, device_map="meta")
 
     if args.model_load_dir and wandb_utils.is_wandb_url(args.model_load_dir):
         wandb_utils.download_model_from_wandb(args.model_load_dir)
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -42,7 +42,7 @@ def main(override_sys_args: Optional[List[str]] = None):
         wandb_utils.download_model_from_wandb(args.model_load_dir)
 
     end = datetime.now()
-    print(f"Weights are downloaded in {end - start} seconds")
+    print(f"Weights downloaded in {end - start} seconds")
 
 
 def raise_on_weights_not_downloaded(model_ids: List[str]):
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -1,3 +1,4 @@
+from typing import List, Optional
 from datetime import datetime
 
 import huggingface_hub
@@ -8,11 +9,11 @@ from ultravox.training import config_base
 ALLOW_PATTERNS = ["*.safetensors", "*.json"]
 
 
-def main():
+def main(args: Optional[List[str]] = None):
     start = datetime.now()
     print("Downloading weights ...")
 
-    args = config_base.get_train_args()
+    args = config_base.get_train_args(args)
 
     for model_id in [args.text_model, args.audio_model]:
         try:
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -24,6 +24,7 @@ def main(override_sys_args: Optional[List[str]] = None):
                 repo_id=model_id, allow_patterns=ALLOW_PATTERNS
             )
             # A backstop to make sure the model is fully downloaded even if ALLOW_PATTERNS is not enough
+            # Using `device_map="meta"` to avoid loading the weights into memory or device
             transformers.AutoModel.from_pretrained(model_id, device_map="meta")
         except huggingface_hub.utils.GatedRepoError as e:
             raise e
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -2,8 +2,8 @@ from datetime import datetime
 
 import huggingface_hub
 
-from ultravox.training import config_base
 from ultravox.model import wandb_utils
+from ultravox.training import config_base
 
 ALLOW_PATTERNS = ["*.safetensors", "*.json"]
 
--- a/ultravox/training/helpers/prefetch_weights.py
+++ b/ultravox/training/helpers/prefetch_weights.py
@@ -9,11 +9,11 @@ from ultravox.training import config_base
 ALLOW_PATTERNS = ["*.safetensors", "*.json"]
 
 
-def main(args: Optional[List[str]] = None):
+def main(override_sys_args: Optional[List[str]] = None):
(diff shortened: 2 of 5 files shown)


farzadab:
@liPatrick I'm gonna add you to the PRs for visibility. You don't have to review.

ultravox/training/train.py
```diff
@@ -166,9 +152,10 @@ def train(args: config_base.TrainConfig):
         logging.info(f"Loading model state dict from {args.model_load_dir}")
         load_path = args.model_load_dir
         if wandb_utils.is_wandb_url(load_path):
-            # Download the model from W&B. The main process should do the download while the others wait.
-            with ddp_utils.run_on_master_first(is_master):
-                load_path = wandb_utils.download_model_from_wandb(load_path)
+            # We assume that the weights are already downloaded via prefetch_weights.py
+            # and hence this is just resolving the path. If the weights are not downloaded,
+            # we might see a race condition here when using DDP.
+            load_path = wandb_utils.download_model_from_wandb(load_path)
```
juberti, line 158: Is it worth changing this API to essentially only resolve the path then, to avoid errors?

farzadab: Yeah I was thinking of that. I don't remember seeing that option in W&B API before. I will take another look.
I was thinking that keeping it won't hurt since it still supports non-DDP usages, specially if someone forgets to do prefetch_weights.
Also, IIRC W&B generally did a good job handling the race conditions even in DDP (we had mainly added `run_on_master_first` for HF downloads).

ultravox/training/helpers/prefetch_weights.py
```diff
@@ -0,0 +1,38 @@
...
+
+def main(override_sys_args: Optional[List[str]] = None):
+    start = datetime.now()
+    print("Downloading weights ...")
+
+    args = config_base.get_train_args(override_sys_args)
+
+    for model_id in [args.text_model, args.audio_model]:
+        try:
+            huggingface_hub.snapshot_download(
```
farzadab, line 20: Also I forgot to mention: the reason for using `snapshot_download` over `AutoModel.from_pretrained` is that the download happens in parallel (~5 mins for the whole 70B).

I also tried [`hf-transfer`](https://github.com/huggingface/hf_transfer) that @petersalas mentioned, but even though each file was downloaded faster that way, no parallel download was happening so this one seemed to be faster overall in my tests.

ultravox/training/train.py
```diff
@@ -129,10 +116,9 @@ def train(args: config_base.TrainConfig):
 
     logging.info("Instantiating model...")
 
-    # Since the model downloads the language model and audio encoder weights, we want one process to finish up
-    # downloading before the others start in order to avoid race conditions.
-    with ddp_utils.run_on_master_first(is_master):
-        model = ultravox_model.UltravoxModel(config)
+    # We assume that the weights are already downloaded via prefetch_weights.py
+    # If the weights are not downloaded, we might see a race condition here when using DDP.
+    model = ultravox_model.UltravoxModel(config)
```
farzadab, line 121: We could still keep `run_on_master_first` here to be safe (since we manually only download the `.safetensors` files and that can potentially be insufficient). But since that would've made the start time slower, I decided the added safety wasn't worth it.

farzadab: Actually never mind, I added a `.from_pretrained` call to cover that case too.

ultravox/training/train.py
farzadab: I can see a `self._default_root()` to resolve the path but I'd rather not rely on a private method.
I also see a `should_download_entry(entry)`, but that also requires looping over contents.

I don't see an option that I would consider better I think.

mcloud.yaml
```diff
@@ -10,7 +10,7 @@ integrations:
     git_branch: $UV_BRANCH
     pip_install: poetry==1.7.1
 command: >-
-  cd ultravox && poetry install --no-dev && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
+  cd ultravox && poetry install --no-dev && poetry run python -m ultravox.training.helpers.prefetch_weights $TRAIN_ARGS && poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
```
zqhuang211, line 13: Can we break this command into multiple lines for improved readability:

```
 cd ultravox &&
 poetry install --no-dev &&
 poetry run python -m ultravox.training.helpers.prefetch_weights $TRAIN_ARGS &&
 poetry run torchrun --nproc_per_node=8 -m ultravox.training.train $TRAIN_ARGS
 ```

File: ultravox/training/helpers/prefetch_weights.py [hunk 5]
```diff
@@ -0,0 +1,44 @@
...
+            )
+            # A backstop to make sure the model is fully downloaded even if ALLOW_PATTERNS is not enough
+            # Using `device_map="meta"` to avoid loading the weights into memory or device
+            transformers.AutoModel.from_pretrained(model_id, device_map="meta")
+        except huggingface_hub.utils.GatedRepoError as e:
+            raise e
+        except huggingface_hub.utils.RepositoryNotFoundError as e:
+            print(
+                f"Model {args.text_model} not found on HF Hub. Skipping download. Error: {e}"
+            )
```
zqhuang211, line 34: So it is ignored if the repository is not found?

ultravox/training/train.py
```diff
@@ -129,10 +116,9 @@ def train(args: config_base.TrainConfig):
 
     logging.info("Instantiating model...")
 
-    # Since the model downloads the language model and audio encoder weights, we want one process to finish up
-    # downloading before the others start in order to avoid race conditions.
-    with ddp_utils.run_on_master_first(is_master):
-        model = ultravox_model.UltravoxModel(config)
+    # We assume that the weights are already downloaded via prefetch_weights.py
```
zqhuang211, line 119: If the weights are already downloaded, it would be ideal to check this condition first and provide an appropriate error message if they are not found.

We could still include the model initialization within `ddp_utils.run_on_master_first(is_master)` to ensure it’s only executed once when using DDP.

ultravox/training/train.py
zqhuang211: I think I miss something here. Why is it necessary to remove the `with ddp_utils.run_on_master_first(is_master)` statement?

ultravox/training/helpers/prefetch_weights.py
farzadab: The assumption is that this is going to be a local model in this case.
I can add a check to see if the directory exists.

ultravox/training/train.py
farzadab: It's not necessary to remove it. There's just no reason for having it and it will make the start time of our training runs slower.

farzadab: The reason we had it in the first place was to avoid race-conditions when downloading the model.

ultravox/training/helpers/prefetch_weights.py
zqhuang211: Got it. Maybe the logic should be:

1. Fetch from HF
2. If repository not found on HF, check if local model exists
3. Fail with an error message

ultravox/training/train.py
zqhuang211: Agreed, `run_on_master_first` would make the other processes wait for no reason.

I think my concern is that we should avoid getting into a race condition. Would something like this (from o1) work?

```
from huggingface_hub import snapshot_download
from huggingface_hub.utils import LocalEntryNotFoundError

model_name = "bert-base-uncased"  # Replace with your model name

try:
    # Attempt to download the model snapshot locally without internet access
    snapshot_download(repo_id=model_name, local_files_only=True)
    print(f"The model '{model_name}' is already downloaded locally.")
except LocalEntryNotFoundError:
    print(f"The model '{model_name}' is not downloaded locally.")
```

ultravox/training/helpers/prefetch_weights.py
farzadab: I moved the `.from_pretrained` call outside of try/catch and updated comments. That should handle this case as well.

ultravox/training/train.py
farzadab: I'm confused. What scenario are you trying to cover?

farzadab: And no this code doesn't fix the issue if the files are not downloaded already.

zqhuang211: If I understand correctly, the training procedure works in two ways:

1. When training on a single process, it doesn’t matter whether we run prefetch_weights or not.
2. When training on multiple processes, prefetch_weights should be run separately before the training run starts.

I think this distinction is fine. However, we need a mechanism to ensure that in the second case, prefetch_weights is executed first, and if not, it should be detected.

farzadab: There is no race-condition on a single process. There's only one process and it's going to download things normally.

farzadab: It is going to get caught. Again, are you just arguing for improving the error message you get?

mcloud.yaml
farzadab: Worked out of the box. Nice!

ultravox/training/train.py
farzadab: I added an imperfect check. I still don't think the extra code is useful.

ultravox/training/helpers/prefetch_weights.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # A backstop to make sure the model is fully downloaded. Scenarios to consider:
+        # - ALLOW_PATTERNS is not enough to download all files needed
+        # - The model is local, this will verify that everything is in order
+        # Using `device_map="meta"` to avoid loading the weights into memory or device
+        transformers.AutoModel.from_pretrained(model_id, device_map="meta")
+
+    if args.model_load_dir and wandb_utils.is_wandb_url(args.model_load_dir):
```
juberti, line 41: Should we skip the HF download above if it's a wandb URL?

ultravox/training/helpers/prefetch_weights.py
```diff
@@ -0,0 +1,60 @@
...
+        wandb_utils.download_model_from_wandb(args.model_load_dir)
+
+    end = datetime.now()
+    print(f"Weights are downloaded in {end - start} seconds")
```
juberti, line 45:
```suggestion
    print(f"Weights downloaded in {end - start} seconds")
```

ultravox/training/train.py
```diff
@@ -129,10 +118,18 @@ def train(args: config_base.TrainConfig):
 
     logging.info("Instantiating model...")
 
-    # Since the model downloads the language model and audio encoder weights, we want one process to finish up
-    # downloading before the others start in order to avoid race conditions.
-    with ddp_utils.run_on_master_first(is_master):
-        model = ultravox_model.UltravoxModel(config)
+    if is_distributed:
```
juberti, line 121: Why do this only when we're running on multi-GPU?

juberti (approved):
LG, just nits.

ultravox/training/helpers/prefetch_weights.py
farzadab: Right now we don't have the option of training audio/text model fully, so the W&B models are always incomplete without the base model.

ultravox/training/train.py
farzadab: Short answer: on a single process there is no race condition, so there's no reason to force that `prefetch_weights.py` was run.

TBH I'm conflicted about whether we even need to enforce this, why raise an error only in anticipation of something going wrong. I added it because @zqhuang211 asked for it.

But the idea is that there are two scenarios:
1. if there's only one process, running `train.py` without `prefetch_weights.py` is fine. No reason to panic.
2. if there are multiple processes, we are likely to get an error, so we'll have to run prefetch_weights. The point of this code is to make the error message more easily understandable I guess

farzadab: A very minor nit: it's possible to have multiple GPUs and still have one process (as in DataParallel).

ultravox/training/train.py
farzadab: Moved `raise_on_weights_not_downloaded` above `config` and `tokenizer` initialization in light of the new failure mode being discovered.

ultravox/training/train.py
farzadab: I added a prefetch_weights in train.py too to make local/test runs easier.
I'm not seeing any response on this comment. So I'm gonna assume it's resolved.

PR #107, opened 2024-09-10
[70B-Part2] Improved save model (that can work with FSDP)

farzadab:
The fixes in this PR might not make much sense on their own, but here's what's changing:

1. no_split_modules is getting set dynamically. Apparently the previous approach (of including all possible classes) leads to an error since all classes are expected to be present.

2. `state_dict` and `load_state_dict` logic are slightly modified in how they're applied.
  a. using existing register_hook method instead
  b. changing `save_pretrained` instead of `state_dict`. TODO: This might end up fixing some of the warnings we were seeing and suppressing as well (not tested yet).
  
3. `train.py` reverts to using `trainer.save_model` instead of pipeline (in order to work with FSDP), but we will still save the pipeline code and configs.

### DIFF
--- a/ultravox/model/ultravox_model.py
+++ b/ultravox/model/ultravox_model.py
@@ -34,7 +34,6 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
 
     config_class = UltravoxConfig
     config: UltravoxConfig  # for type hinting
-    _no_split_modules = ["Wav2Vec2Model", "WhisperEncoder", "LlamaDecoderLayer"]
     # We minimize the weights in state_dict in order to reduce the size of the checkpoint
     # The issue is that load_pretrained() uses state_dict() keys to know what keys are expected
     # As such we have to tell is to ignore some keys that are not always in the model
@@ -46,6 +45,7 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
 
     def __init__(self, config: UltravoxConfig):
         super().__init__(config)
+        self._register_load_state_dict_pre_hook(self._pre_load_state_dict_hook)
 
         self.keep_params: Set[str] = set()
         self.vocab_size = config.vocab_size
@@ -53,6 +53,13 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
         self.audio_tower = self._create_audio_tower(config)
         self.multi_modal_projector = UltravoxProjector(config)
         self.language_model = self._create_language_model(config)
+
+        # Determine no_split_modules dynamically to use with FSDP auto_wrap policy.
+        # FSDP throws an error if some of the layer types are not found in the model.
+        # This would be something like ["LlamaDecoderLayer", "WhisperEncoderLayer"]
+        self._no_split_modules = (
+            self.language_model._no_split_modules + self.audio_tower._no_split_modules
+        )
 
         self.loss_config = LossConfig()
         self.post_init()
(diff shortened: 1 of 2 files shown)


ultravox/training/train.py
```diff
@@ -317,12 +317,17 @@ def train(args: config_base.TrainConfig):
...
-        pipeline.save_pretrained(args.output_dir)
+    # Save the pipeline code and update the config to include the pipeline
+    pipeline = ultravox_pipeline.UltravoxPipeline(
+        model, tokenizer=text_tokenizer, device=model.device
+    )
+    # We don't want to save the model twice. Trainer.save_model saves the model to the output_dir.
+    old_save_pretrained = model.save_pretrained
+    model.save_pretrained = lambda *_, **__: None
+    pipeline.save_pretrained(args.output_dir)
+    model.save_pretrained = old_save_pretrained
```
farzadab, line 328: This part is quite hacky, I agree. I didn't know how else to achieve the same result without making our save process even slower.
Note that this part of the code will save both the pipeline code as well as its config, both of which are required for `oaieval` integration to work.

ultravox/model/ultravox_model.py
```diff
@@ -510,6 +516,7 @@ class ModifiedWhisperEncoder(whisper.WhisperEncoder):
     """
 
     base_model_prefix = "model.encoder"
+    _no_split_modules = ["WhisperEncoderLayer"]
```
juberti, line 519: For my own knowledge, this means Whisper won't be sharded across GPUs (since we don't need it to be?)

ultravox/training/train.py
```diff
@@ -317,12 +317,17 @@ def train(args: config_base.TrainConfig):
...
-        # Saving the model using pipeline to ensure its code is saved
-        pipeline = ultravox_pipeline.UltravoxPipeline(
-            model, tokenizer=text_tokenizer, device=device
-        )
-        pipeline.save_pretrained(args.output_dir)
+    # Save the pipeline code and update the config to include the pipeline
+    pipeline = ultravox_pipeline.UltravoxPipeline(
+        model, tokenizer=text_tokenizer, device=model.device
+    )
+    # We don't want to save the model twice. Trainer.save_model saves the model to the output_dir.
```
juberti, line 324: But why is a standalone `save_model` call needed at all?

farzadab: Because `save_model` knows how to save FSDP models correctly but pipeline does not (see attached pieces of code).
You could argue that the same might be replicable with pipeline given the right context object and that would be fair. I finally got something working and didn't want to touch it 😅

I can take another pass at it.

image
image

ultravox/model/ultravox_model.py
farzadab: Actually my understanding is that it will be.
It's just saying where to shard: at the encoder-layer level for Whisper and at the decoder-layer level in Llama. So each layer is on one GPU and you don't split further.

I believe it's also saying the layers that are not part of this (e.g. embedding layers or convs) will not be sharded at all.

ultravox/training/train.py
farzadab: I did some more hacking around and I still don't see a way out of using the `trainer.save_model` call.

The `pipeline` however, is only needed for saving the pipeline code and updating the config.
I could copy and paste the following code, which would look nicer in the short-term, but I feel like it's safer not to do that (in case `transformers`' way of doing things changes):
image

farzadab:
@juberti @zqhuang211 are there any more comments?
If not, please approve so we can merge this.

ultravox/training/train.py
juberti: OK, I suggest just clarifying the comment to explain this, ie that the pipeline doesn't know how to save FSDP models properly, so we need to do it manually.

juberti (approved):

PR #108, opened 2024-09-10
Not uploading text_config when text_model_id is present

farzadab:
As the title suggests, removing `text_config` when `text_model_id` is present, since `text_config` will be overwritten and its existence can be confusing.

This PR is in partial response to this comment on Discord by Hyunsung Lee:
image

I should note that this PR doesn't quite address this comment, but simply improves another confusion that can arise later down the line when people try to update text_config but see it doesn't actually do anything.

Example config (revision of a private repo): https://huggingface.co/fixie-ai/ultravox-v0_4-llama-3_1-70b/blob/4cfd42ee20b5c2df2065d25a223ff10bfaf0d728/config.json
image

### DIFF
--- a/ultravox/model/ultravox_config.py
+++ b/ultravox/model/ultravox_config.py
@@ -99,7 +99,6 @@ class UltravoxConfig(transformers.PretrainedConfig):
         audio_model_id: Optional[str] = None,
         text_model_id: Optional[str] = None,
         ignore_index: int = -100,
-        audio_token_index: int = 32000,
         hidden_size: int = 4096,
         stack_factor: int = 8,
         norm_init: float = 0.4,
@@ -112,7 +111,6 @@ class UltravoxConfig(transformers.PretrainedConfig):
 
         self.audio_model_id = audio_model_id
         self.text_model_id = text_model_id
-        self.audio_token_index = audio_token_index
 
         self.hidden_size = hidden_size
         self.stack_factor = stack_factor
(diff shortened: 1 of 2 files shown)


ultravox/model/ultravox_config.py
```diff
@@ -99,7 +99,6 @@ def __init__(
         audio_model_id: Optional[str] = None,
         text_model_id: Optional[str] = None,
         ignore_index: int = -100,
-        audio_token_index: int = 32000,
```
farzadab, line 102: We have not been using `audio_token_index` for a while now.

ultravox/model/ultravox_config_test.py
```diff
@@ -0,0 +1,37 @@
...
+import transformers
+
+from ultravox.model import ultravox_config
+
+
+@pytest.mark.parametrize(
+    "model_id",
+    ["fixie-ai/ultravox-v0_2", "fixie-ai/ultravox-v0_3", "fixie-ai/ultravox-v0_4"],
+)
+def test_can_load_release(model_id: str):
```
farzadab, line 11: I think this could be a good way to make sure we stay backward compatible. If there's a breaking change needed, then we can consciously decide if we want to go ahead with it.

It obviously won't cover everything, but it should be a good start.

ultravox/model/ultravox_config.py
farzadab: Also, as the tests also verify, removing this will not cause issues when `audio_token_index` is actually present in `config.json`.

juberti (approved):

ultravox/model/ultravox_config.py
petersalas: As the Mistral experimentation uncovered, `audio_token_index` or something like it would be very useful downstream to have a token ID that is sensible to use as a placeholder with the particular tokenizer. In vLLM we currently hardcode the "<|reserved_special_token_0|>" Llama token, which broke when I tried to use a Mistral LLM.

petersalas: That said, we don't currently use this in vLLM, so removing it won't break anything to my knowledge.

farzadab: Okay good to know. I'll keep it then.

farzadab: Actually I'll remove it for now and then we'll add it back in coordination with vllm, because the current value (32000) is probably just plain wrong.

PR #110, opened 2024-09-11
Support longer audio contexts

liPatrick:
The whisper encoder has a max context of 30s of audio. 

This pr enables our model to support longer contexts by splitting long audio into chunks of 30sec (with the last chunk being the exception)

Note* should also work in batch mode.

### DIFF
--- a/ultravox/inference/infer_test.py
+++ b/ultravox/inference/infer_test.py
@@ -60,7 +60,7 @@ class FakeInference(infer.LocalInference):
         )
         self.model.device = "cpu"
         self.model.generate = mock.MagicMock(side_effect=fake_generate)
-        self.model.audio_tower_context_length = 3000
+        self.model.audio_tower_context_length = None
 
 
 EXPECTED_TOKEN_IDS_START = [128000, 128006, 882, 128007]
--- a/ultravox/inference/infer_test.py
+++ b/ultravox/inference/infer_test.py
@@ -49,7 +49,7 @@ class FakeInference(infer.LocalInference):
             return output
 
         processor = ultravox_processing.UltravoxProcessor(
-            audio_processor, tokenizer=tokenizer
+            audio_processor, tokenizer=tokenizer, audio_context_size=None
         )
         super().__init__(
             mock.MagicMock(),
@@ -60,7 +60,6 @@ class FakeInference(infer.LocalInference):
         )
         self.model.device = "cpu"
         self.model.generate = mock.MagicMock(side_effect=fake_generate)
-        self.model.audio_tower_context_length = None
 
 
 EXPECTED_TOKEN_IDS_START = [128000, 128006, 882, 128007]
--- a/ultravox/inference/infer_test.py
+++ b/ultravox/inference/infer_test.py
@@ -60,6 +60,7 @@ class FakeInference(infer.LocalInference):
         )
         self.model.device = "cpu"
         self.model.generate = mock.MagicMock(side_effect=fake_generate)
+        self.model.audio_tower_context_length = 3000
 
 
 EXPECTED_TOKEN_IDS_START = [128000, 128006, 882, 128007]
--- a/ultravox/inference/infer_test.py
+++ b/ultravox/inference/infer_test.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from unittest import mock
 
 import numpy as np
@@ -30,16 +31,23 @@ def audio_processor():
     )
 
 
+@pytest.fixture(scope="module")
+def audio_processor_whisper():
+    return transformers.AutoProcessor.from_pretrained("openai/whisper-tiny")
+
+
 class FakeInference(infer.LocalInference):
     def __init__(
         self,
         tokenizer: transformers.PreTrainedTokenizer,
         audio_processor: transformers.ProcessorMixin,
+        audio_context_size: Optional[int] = None,
     ):
         def fake_generate(**kwargs):
             input = kwargs.get("input_ids")
+            input_len = input.shape[1] if input is not None else 0
             output = transformers.generation.utils.GenerateDecoderOnlyOutput(
-                sequences=[range(25)]
+                sequences=[range(input_len + 5)]  # Always output 5 tokens
             )
             streamer = kwargs.get("streamer", None)
             if streamer:
@@ -49,7 +57,7 @@ class FakeInference(infer.LocalInference):
             return output
 
         processor = ultravox_processing.UltravoxProcessor(
-            audio_processor, tokenizer=tokenizer, audio_context_size=None
+            audio_processor, tokenizer=tokenizer, audio_context_size=audio_context_size
         )
         super().__init__(
             mock.MagicMock(),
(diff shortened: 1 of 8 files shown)


ultravox/model/ultravox_model.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, config: UltravoxConfig):
         self.vocab_size = config.vocab_size
 
         self.audio_tower = self._create_audio_tower(config)
+        self.audio_tower_context_length = (
+            3000  # the context window for the whisper model
+        )
```
farzadab, line 56: We don't want Ultravox to only work with Whisper.

ultravox/model/ultravox_model.py
```diff
@@ -186,17 +190,29 @@ def forward(
...
+            # Check if the audio_values T dimension is greater than whisper encoder's context window.
+            if audio_values.shape[2] > self.audio_tower_context_length:
+                audio_values_chunks = torch.split(
+                    audio_values, self.audio_tower_context_length, dim=2
+                )
+            else:
+                audio_values_chunks = (audio_values,)
+
+            rebuilt_audio_embeds = []
+            for audio_chunk in audio_values_chunks:
```
farzadab, line 202: Doing a for loop is generally not a good idea if the operation can be vectorized.

farzadab (commented):
Thanks Pat.

Your approach is valid, but I had a slightly different take on how we would want to approach this so that it would later on allow for having multiple (zero or more) audio tracks per sample.

Let's meet if the explanation I gave is vague.

ultravox/model/ultravox_model.py
```diff
@@ -186,17 +190,29 @@ def forward(
...
-            # B x A/3200 x D
-            audio_tower_output = self.audio_tower.forward(
-                audio_values
-            ).last_hidden_state
-            audio_tower_output = audio_tower_output.to(inputs_embeds.dtype)
+            # Check if the audio_values T dimension is greater than whisper encoder's context window.
+            if audio_values.shape[2] > self.audio_tower_context_length:
+                audio_values_chunks = torch.split(
+                    audio_values, self.audio_tower_context_length, dim=2
+                )
```
farzadab, line 197: Huh, this message was removed somehow. You must've been so confused!

Okay so one approach I think we could take to handle both problems is:
- handle breaking up the audio in processor instead (it might already have some methods for doing so, not a big deal if not either)
- now we can't assume audio_embeds and input_embeds are matched in their index. The collator should make sure to correctly index all the audio chunks and provide the index to the forward function
- the forward function doesn't need to change much, just that it takes the `audio_batch_index` as an input and uses that to know where to put the audio_embeds instead of assuming the indices match

ultravox/model/ultravox_model.py
zqhuang211: Agreed, this length restriction should only apply to the whisper encoder

ultravox/model/ultravox_model.py
zqhuang211: I agree this is not the right place to handle this.

The length restriction of the whisper encoder should be handled in the forward method of the whisper encoder.

farzadab: I don't think that's how Whisper works (e.g. see how [ASR pipeline](https://github.com/huggingface/transformers/blob/a05ce550bf1cda8ae3bffcc62607597770460a42/src/transformers/pipelines/automatic_speech_recognition.py#L350) handles this). It relies on processor to create the right-sized chunks.

From what I've seen in `transformers`, usually things like truncation, padding, and splitting are handled by processor to allow for better configurability and to not make the model too complicated.

farzadab: Another reason why doing this in forward function is bad (be it Ultravox.forward or Whisper.forward), is that padding happens first and then things get chunked.

That means if you have one sample that is too big in a batch, all of your samples will have a lot of chunks.

zqhuang211: I think you are right, it should be first split in the processor. Let's discuss offline.

liPatrick:
Results from eval on long audio contexts: 
https://wandb.ai/fixie/ultravox/runs/jz9uhw4e/overview

juberti:
Interesting. Do you have a sample output dataset I could take a look at?

liPatrick:
> Interesting. Do you have a sample output dataset I could take a look at?

yeah, take a look here for the original dataset: https://huggingface.co/datasets/fixie-ai/covost2_long_audio 

here for the output results: https://wandb.ai/fixie/ultravox/runs/2qtpd764/files/output

juberti:
Hmm, I listened to a few clips and I wonder if the merging is the right way to do this. The audio clips tend to be fairly different with their own speaking rate and volume level, and the combined audio just doesn't make much sense (e.g., ambiguous pronoun resolution), which perhaps explains the hit to the es-en ASR metrics.

I wonder if we could use something like librispeech and stitch some of the segmented clips back together so they are coherent. Or maybe just find an ASR dataset with longer samples - we probably don't need a ton of data here.

zqhuang211:
> I wonder if we could use something like librispeech and stitch some of the segmented clips back together so they are coherent. Or maybe just find an ASR dataset with longer samples - we probably don't need a ton of data here.

Yes, some speech datasets, like LibriSpeech and GigaSpeech, have meta-information that supports this. I kind of prefer stitching back segments instead of selecting longer samples so that you have a comparable baseline (segmented) with the former approach. We could find similar datasets for speech translation too.

The synthesized data is noisier but also harder. If the model is able to perform at the human level, there shouldn’t be any significant difference in performance if it transcribes/translates an individual segment at a time or five segments at a time.

The approach you suggested would produce more realistic datasets and hopefully can provide a measure of how the model benefits from longer context.

I think both are valid and measure different aspects of the model.

juberti:
OK, I can get behind that. I still think this warrants further investigation though:
```
eval/covost2-asr-es_en.2k-asr:0.14595425715933116
eval/covost2_long_audio-asr-combine-5-es_en.2k-asr:0.1345223909283106
eval/covost2_long_audio-asr-combine-10-es_en.2k-asr:0.20847972323659428
```
It seems odd that combining 5x is much better than combining 10x, and suggests a potential problem in the merging.

Also, did you run any numbers for the duplicate (rather than merge) approach?

liPatrick:
Yeah I ran some evals for the duplicate, and what ended up happening was it would only transcribe the first utterance. So the WER was really high (around 0.82).

juberti:
Hmm, that's kind of surprising. Perhaps the encoder hidden states end up the same when repeated audio is used?

ultravox/inference/infer_test.py
```diff
@@ -60,6 +60,7 @@ def fake_generate(**kwargs):
         )
         self.model.device = "cpu"
         self.model.generate = mock.MagicMock(side_effect=fake_generate)
+        self.model.audio_tower_context_length = None
 
 
 EXPECTED_TOKEN_IDS_START = [128000, 128006, 882, 128007]
```
juberti, line 66: suggest adding a new test to this file to demonstrate the long-audio-context handling in the processor

ultravox/model/ultravox_processing.py
```diff
@@ -89,6 +90,7 @@ def __call__(
         text: Optional[str] = None,
         audio: Optional[Union[np.ndarray, torch.Tensor]] = None,
         sampling_rate: Optional[int] = None,
+        audio_context_size: Optional[int] = None,
```
farzadab, line 93: I would probably put `audio_context_size` in the processor `__init__` rather than `__call__` because it's not expected to change with each sample.

Also I feel like `UltravoxProcessor` class should come up with the right defaults on its own, otherwise you'd have to specify it in too many places: 1) inference (you already did)  2) training  3) when someone else loads UltravoxPipeline via `.from_pretrained`.

PR #111, opened 2024-09-12
Unified dataset specification

zqhuang211:
This PR unifies the dataset configuration and training/validation/evaluation specifications for datasets.

The datasets used for training, validation, and evaluation are configured in the `train_dataset_configs`, `val_dataset_configs`, and `eval_dataset_configs` fields using a config YAML file, with global settings specified in `train_dataset_args`, `val_dataset_args`, and `eval_dataset_args`. Most datasets (including those used in the v0.4 release) can be configured using the GenericVoiceDataset class, while legacy dataset classes can be configured similarly, with the default Huggingface/MDS paths hardcoded. This approach offers several benefits:

1. Unified support for both legacy and new datasets.
2. Simplified configuration for data splits, subsets, and weighting across all datasets.
3. Greater flexibility for experimentation to support different training, validation, and evaluation configurations.

Additionally, the evaluation code has been updated to support batch inference. It is now integrated into the training workflow and supports standalone evaluation.


ultravox/evaluation/eval_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+
+def evaluate_answers(
+    samples: List[eval_types.Sample], metric_config: eval_types.EvalConfig
+) -> eval_types.Result:
+    if metric_config.metric in CORPUS_METRIC_REGISTRY:
+        return CORPUS_METRIC_REGISTRY[metric_config.metric](
+            samples, **metric_config.args
+        )
+    else:
+        raise ValueError(f"Unknown metric: {metric_config.metric}")
```
farzadab, line 35: Why not just do a for loop here instead of refusing the request?

ultravox/evaluation/eval_types.py
```diff
@@ -1,15 +1,27 @@
...
 
 
 @dataclasses.dataclass
 class Sample(dataclasses_json.DataClassJsonMixin):
+    index: int  # index of the sample in the dataset, used for preserving order after ddp all_gather
     question: str
-    generated_answer: str
-    expected_answer: str
-    history: List[Dict[str, str]] = dataclasses.field(default_factory=list)
+    hypothesis: str
```
farzadab, line 22: `hypothesis` is very vague. Why change the name?

ultravox/evaluation/eval_types.py
```diff
@@ -1,15 +1,27 @@
 import dataclasses
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import dataclasses_json
+from pydantic import BaseModel
```
farzadab, line 5: Why use `pydantic`? Everything else in the repo uses a `dataclass`.

ultravox/tools/eval_tool.py
```diff
@@ -1,35 +1,78 @@
...
-        scores.append(result.score)
-        average = sum(scores) / len(scores)
-        print(f"{i}: score={result.score:.2f} average={average:.2f}")
-        if args.verbose and isinstance(result, eval_types.InstructResult):
-            print(f"  reason={result.reason}")
+    world_size = device_helpers.get_world_size()
+    local_rank = device_helpers.get_local_rank()
+    device = torch.device(args.device, index=local_rank)
+    if world_size > 1:
+        dist.init_process_group(backend="gloo")
```
farzadab, line 33: Why use `gloo`?
image

ultravox/tools/eval_tool.py
```diff
@@ -1,35 +1,78 @@
...
+
+    if local_rank == 0:
+        args.output_dir.mkdir(parents=True, exist_ok=True)
+        if "wandb" in args.report_logs_to:
+            wandb.init(
+                project=os.getenv("WANDB_PROJECT", "ultravox"),
+                config=dataclasses.asdict(args),
+                name=args.exp_name,
+                dir="runs",
+            )
```
farzadab, line 43: Nit: A suggestion is to add the option to set the `id` field so that the metrics could be reported on the original W&B run.

ultravox/tools/eval_tool.py
```diff
@@ -1,35 +1,78 @@
...
-    metric: str = simple_parsing.field(default="asr", alias="-m")
-    # Verbose output
-    verbose: bool = simple_parsing.field(default=False, alias="-v")
 
+def main():
+    args = simple_parsing.parse(
+        config_class=config_base.TrainConfig,
+        add_config_path_arg=True,
+        args=[string_helpers.fix_hyphens(arg) for arg in sys.argv[1:]],
+    )
```
farzadab, line 27: nit:
```suggestion
    args = config_base.get_train_args()
```

ultravox/training/config_base.py
```diff
@@ -2,115 +2,189 @@
...
-    # we convert these dictionaries to DataDictConfig objects using Pydantic
-    # to enforce type constraints and validation, in the __post_init__ method.
-    data_dicts: Optional[List[Dict[str, Any]]] = None
+    """Configuration class for training settings."""
+
+    text_model: Optional[str] = None
+    """Language model to use; could be a huggingface model id, wandb path, or local path."""
+
+    audio_model: Optional[str] = None
+    """Audio encoder model to use; could be a huggingface model id, wandb path, or local path."""
```
farzadab, line 25: Why use the doc style instead of the comment style?
I find the comments that come after quite confusing.
`simple_parsing` can use both.

ultravox/training/config_base.py
```diff
@@ -2,115 +2,189 @@
...
+            for data_dict in self.val_dataset_configs
+        ]
+        self.eval_dataset_configs = [
+            datasets.DatasetConfig(**data_dict)
+            for data_dict in self.eval_dataset_configs
+        ]
 
         assert self.data_type in ["bfloat16", "float16", "float32"]
-        if self.device == "cuda" and not torch.cuda.is_available():
-            self.device = "mps" if torch.backends.mps.is_available() else "cpu"
```
farzadab, line 113: Why is this logic being removed?

ultravox/utils/string_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+
+def sanitize_name(name: str) -> str:
+    """
+    Sanitize a string to be safely used as a valid filename, folder name, or Python module name.
+
+    Parameters:
+        name (str): The input string to sanitize.
+
+    Returns:
+        str: The sanitized name.
```
farzadab, line 18: I don't follow the logic of this file at all. Did you write this? If not, what is the source?

ultravox/training/config_base.py
```diff
@@ -122,12 +196,16 @@ def __post_init__(self):
...
+                f"Experiment name {self.exp_name} has been sanitized to {sanitized_exp_name} for use as a valid file/directory/module name."
+            )
+            self.exp_name = sanitized_exp_name
+
         if self.output_dir is None:
             self.output_dir = Path("runs") / self.exp_name
 
-        # HF Pipeline gets tripped up if the path has a "." in it
-        self.output_dir = Path(str(self.output_dir).replace(".", "--"))
-
```
farzadab, line 130: Why is this logic removed?

ultravox/training/configs/llama_whisper.yaml
```diff
@@ -0,0 +1,102 @@
...
+exp_name: "llama3.1-8b-whisper"
+text_model: "meta-llama/Meta-Llama-3.1-8B-Instruct"
+audio_model: "openai/whisper-medium"
+
+loss_config:
+  loss_function: "KL_Divergence"
+
+max_steps: 20 # x8x24 = 2,764,800
+
+train_dataset_args:
```
farzadab, line 12: Why are datasets being rewritten here? Does the llama_whisper combination require a specific set of train/val/eval datasets that other architectures cannot use?

ultravox/training/configs/meta_config.yaml
```diff
@@ -25,7 +17,185 @@ max_steps: 10_000
 save_steps: 0.25
 logging_steps: 100
 
-batch_size: 4
-data_type: "bfloat16"
-
 report_logs_to: ["tensorboard", "wandb"]
+
+# include speech translation (zh-en, en-zh) and transcription (peoples_speech) for validation, with and without audio
+val_dataset_args:
```
farzadab, line 23: Same issue here. What's the point of so much copy & pasting?

ultravox/training/configs/tinyllama_whisper.yaml
```diff
@@ -0,0 +1,102 @@
...
+exp_name: "tinyllama-whisper"
+text_model: "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
+audio_model: "openai/whisper-small"
+
+loss_config:
+  loss_function: "KL_Divergence"
+
+max_steps: 20 # x8x24 = 2,764,800
+
+train_dataset_args:
```
farzadab, line 12: Same

ultravox/training/ddp_utils.py
```diff
@@ -43,3 +47,18 @@ def sharded_iterator(ds: data.IterableDataset, num_shards: int, shard_index: int
...
+    batch = []
+    for idx, sample in enumerate(ds):
+        if idx % num_shards == shard_index:
+            batch.append((idx, sample))
+            if len(batch) == batch_size:
+                yield batch
+                batch = []
+    # Yield any remaining samples in the last incomplete batch
+    if batch:
+        yield batch
```
farzadab, line 64: Why copy & paste this? The same functionality can be used when `batch_size` equals 1. Also I believe the only reference to `sharded_iterator` is already removed.

ultravox/training/helpers/prefetch_weights.py
```diff
@@ -22,8 +22,12 @@ def main(override_sys_args: Optional[List[str]] = None):
     print(f"Weights downloaded in {end - start} seconds")
 
 
-def download_weights(model_ids: List[str], model_load_dir: Optional[str] = None):
+def download_weights(
+    model_ids: List[Optional[str]], model_load_dir: Optional[str] = None
+):
     for model_id in model_ids:
+        if model_id is None:
+            continue
```
farzadab, line 30: What's the use case for making this optional?

farzadab (commented):
Here are some comments. I didn't get to review the full PR.

juberti:
Hi @zqhuang211, there are a lot of interesting ideas in this PR but it has far too much surface area. The key focus here should be on the data/ changes for the generic dataset approach - and we should handle any other changes (e.g., the new eval approach in separate PRs).

For the data/ classes, I agree with Farzad that the configs have a lot of duplication and errors can easily creep in as a result. To me, the benefit of this new approach seems to be that you could add a new dataset entirely via config, with no code changes, but I don't think we want to have to write more config code than the existing amount of Python code. (My prior here is that having to manually handle each language in Covost via config is more complicated than the current setup.)

This also suggests that starting with a smaller PR will be better as we can more easily discuss the relative merits of the approaches when the amount of changes is smaller.

zqhuang211:
> Hi @zqhuang211, there are a lot of interesting ideas in this PR but it has far too much surface area. The key focus here should be on the data/ changes for the generic dataset approach - and we should handle any other changes (e.g., the new eval approach in separate PRs).

Yes, it touches many areas because dataset specification is used throughout different parts of the code. Changes to the dataset specification breaks other parts that rely on the existing methods. The previous GenericDatasetClass is a trade-off that attempts to add a new dataset class that fits within the existing code, but it definitely has its limits. We may need a more comprehensive process to decide what we want to achieve and how to get there as a group.

> For the data/ classes, I agree with Farzad that the configs have a lot of duplication and errors can easily creep in as a result. To me, the benefit of this new approach seems to be that you could add a new dataset entirely via config, with no code changes, but I don't think we want to have to write more config code than the existing amount of Python code. (My prior here is that having to manually handle each language in Covost via config is more complicated than the current setup.)

I agree that some of the duplication in the configs can be removed/reduced (e.g., certain prompts), and these concerns can be addressed.

The config approach supports adding new datasets without adding new code, but this is not the main reason to advocate for the use of configs. I agree that for this use case, there is not a significant difference between using Python code or config, because it would involve a similar amount of work. 

The main reason for using configs is the flexibility to change all aspects of datasets without changing the Python code, e.g.,

1. Which datasets to use (old or new) for training, eval, and validation.
2. What prompt to use for each dataset.
3. Using the same dataset with different prompts.
4. Using different subsets from individual datasets.
4. Changing specifications of different fields of a dataset.

The current approach already exposes some options to the config, e.g., which existing datasets to use and some global settings for how to use the datasets, but it doesn’t allow fine control of individual datasets.

I think if we want to make it flexible (easier and less likely to break python code) to run experiments, we want to expose these controls to the config, so that the same Python code can be reused and continuously improved (and can be made more robust by including additional validation and safeguards).

We could also include some of the configs into defaults, like these in the meta_config.yaml, for example. 

I think it comes down to what we want to change in order to run new validations of experiments, Python code or YAML config. My preference is to minimize changes to Python code, if changes can be made in a well-structured config file. In the end, the use of datasets should be configurable, similar to how model parameters and training specifications are configurable.

zqhuang211:
I want to add that this is not a Python code vs. YAML config problem. The real issue (let’s ignore the global train/eval/val configs for now) is Python code + <dataset_namestring_only> config vs. Python code + YAML config. The use of YAML config is simply more powerful; it can support different levels of control with a unified dataset specification schema and forces us to think modularly about dataset design.

Let’s consider two extremes:

In one extreme, the details of all datasets are hardcoded in Python code, so
```
# YAML config
train_datasets:
  - type: 'librispeech'
  - type: 'commonvoice'
```
is equivalent to
```
# dataset_namestring_only config
train_datasets: ['librispeech', 'commonvoice'] 
```

In the other extreme, a new dataset can be specified using the YAML config following the dataset specification schema, without writing any new lines of code.

The Python code + YAML config approach allows us to use all spectra between these two extremes, depending on different stages of model development and the specifics of datasets:

- Individual datasets hardcoded entirely in Python code, with their own logic, only configurable via the type field in the YAML config, similar to the current design.
- Individual datasets with certain attributes hardcoded in the Python code (e.g., HF path, name, splits, etc.), while the rest of the attributes are configurable via the YAML config.
- A group of datasets with their own shared logic hardcoded in Python code, with some (or all) attributes configurable in the YAML config. GenericVoiceDataset is one example. I can imagine that new groups of datasets will be added in a similar way in the future.

As we go through different stages of model development, if a dataset deserves special treatment, it can be moved from being entirely configurable in YAML to being partially or entirely hardcoded in Python, following this unified approach. Similarly, a hardcoded dataset can become entirely configurable in YAML if we want to share its logic with some other datasets.

Of course, we can include and update the default validation/evaluation datasets in the meta_config.yaml.

I can think of two downsides to this approach:

1. You have to type five more characters, “type:”, for the entirely hardcoded datasets.
2. There is potential for abuse, which in most cases means we need to improve the Python code and may add new features.

PR #113, opened 2024-09-13
Extend audio ds_tool

liPatrick:
Adding an extend audio task in ds_tool to create longer audio segments for eval

### DIFF
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -446,7 +446,11 @@ class DatasetChunkProcessor:
             "split": split_name,
         }
         assert isinstance(self.args.upload_name, str)
-        ds_split_chunked.push_to_hub(self.args.upload_name, **hub_args)
+        try: 
+            ds_split_chunked.push_to_hub(self.args.upload_name, **hub_args)
+        except Exception as e:
+            print(f"Failed to upload chunk to hub: {e}")
+            raise e
 
 
 def main(args: DatasetToolArgs):
@@ -478,8 +482,6 @@ def main(args: DatasetToolArgs):
         if args.num_samples:
             ds_split = ds_split.select(range(args.num_samples))
 
-        if args.upload_split:
-            split_name = args.upload_split
         ds_chunk_proc.process_and_upload_split_rescursive(
             split_name, ds_split, 0, len(ds_split)
         )
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -232,6 +232,7 @@ class AudioExtensionTask:
         repeated_sentence = " ".join(sentence)
         repeated_translation = " ".join(translation)
         sample[self.audio_column_name]["array"] = repeated_audio
+        sample[self.audio_column_name].pop("path")
         sample[self.asr_column_name] = repeated_sentence
         sample[self.translation_column_name] = repeated_translation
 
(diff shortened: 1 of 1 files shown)


ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,97 @@ def _map_sample(self, sample, exclude_fields):
         return sample
 
 
+@dataclasses.dataclass
+class AudioExtensionTask:
+    audio_column_name: str = simple_parsing.field(default="audio", alias="-a")
+    asr_column_name: str = simple_parsing.field(default="sentence", alias="-A")
+    translation_column_name: str = simple_parsing.field(
```
juberti, line 183: some datasets (eg librispeech) won't have this. Should this be optional?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,97 @@ def _map_sample(self, sample, exclude_fields):
         return sample
 
 
+@dataclasses.dataclass
+class AudioExtensionTask:
+    audio_column_name: str = simple_parsing.field(default="audio", alias="-a")
+    asr_column_name: str = simple_parsing.field(default="sentence", alias="-A")
```
juberti, line 182: maybe `text_column_name` to mirror `audio_column_name`? This confused me at first.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,97 @@ def _map_sample(self, sample, exclude_fields):
...
+    translation_column_name: str = simple_parsing.field(
+        default="translation", alias="-T"
+    )
+    id_column_name: str = simple_parsing.field(default="id", alias="-i")
+    extend_type: str = simple_parsing.field(
```
juberti, line 187: Consider also "-o" for "operation"

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,97 @@ def _map_sample(self, sample, exclude_fields):
...
+
+        if not isinstance(audio, dict) or "array" not in audio:
+            raise ValueError(f"Unsupported audio format: {type(audio)}")
+
+        audio_data = audio["array"]
+        repeated_audio = np.tile(audio_data, self.multiplier)
+        repeated_sentence = " ".join([sentence] * self.multiplier)
+        repeated_translation = " ".join([translation] * self.multiplier)
+
+        new_audio = {key: value for key, value in audio.items() if key != "path"}
```
juberti, line 235: I think there are only 3 keys in an Audio, and the only one we care about copying is `sampling_rate`. Maybe just init the Audio object directly here?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,97 @@ def _map_sample(self, sample, exclude_fields):
...
+            )
+        else:
+            raise ValueError(f"Unknown extend_type: {self.extend_type}")
+
+    def _map_sample_repeat(self, sample):
+        audio = sample[self.audio_column_name]
+        sentence = sample[self.asr_column_name]
+        translation = sample[self.translation_column_name]
+
+        if not isinstance(audio, dict) or "array" not in audio:
```
juberti, line 227: Might be able to handle this automatically by using `ds_split.cast_column` to `Audio`. (Note also that this doesn't exist in the combine operation below)

juberti (commented):
LG overall, just a few nits.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Hm, actually, i think if array isn't in audio, it'll just throw an error which should be fine in this case. I'm hesitant to stack map operations because it adds a lot of time to the processing for larger datasets.

ultravox/tools/ds_tool/ds_tool.py
liPatrick: Yes, good point. We should be able to process transcription only datasets as well.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +176,103 @@ def _map_sample(self, sample, exclude_fields):
         return sample
 
 
+@dataclasses.dataclass
+class AudioExtensionTask:
+    audio_column_name: str = simple_parsing.field(default="audio", alias="-a")
+    text_column_name: str = simple_parsing.field(default="sentence", alias="-A")
```
juberti, line 182:
```suggestion
    text_column_name: str = simple_parsing.field(default="sentence", alias="-t")
```

juberti (approved):

PR #117, opened 2024-09-17
[70B-Part4] Config and init_empty_weights

farzadab:
The final PR for FSDP/70B training, containing the 70B config.


ultravox/training/configs/llama_70b.yaml
```diff
@@ -0,0 +1,16 @@
+# Training configuration for Ultravox with Meta-Llama-3.1-70B-Instruct and Whisper-Medium
+# This configuration is used alongside release_config.yaml
+
+# TODO: make sure to increase max_duration in mcloud.yaml to 18 hours instead of 6
```
farzadab, line 4: We currently don't have a way of influencing `mcloud.yaml` from the configs. I've thought about adding a wrapper script on top that can also give us more flexibility (e.g. additive names or objects). I don't think the need was great enough at the moment though.

ultravox/training/train.py
```diff
@@ -128,7 +130,17 @@ def train(args: config_base.TrainConfig):
...
+    model_load_context = (
+        accelerate.init_empty_weights()
+        if args.use_fsdp and not is_master
+        else contextlib.nullcontext()
+    )
+    # If we're using FSDP, we can just initialize the model on the main process
+    # and use sync_model_states to distribute the weights to the other processes.
+    # Otherwise we'd be loading the model on every process, which uses too much CPU memory.
+    with model_load_context:
+        model = ultravox_model.UltravoxModel(config)
```
farzadab, line 143: This didn't seem necessary for 70B with our current cloud config, but it could easily become a bottleneck.

The issue: The full model gets loaded on RAM per processor and then is moved to GPU. FSDP offers a different approach via the `sync_model_states` option that propagates the weights from the main process.

I've experimented with and without this setting and the training results are exactly the same. I don't have system metrics before W&B starts however to show the expected difference in RAM usage.

ultravox/training/configs/llama_70b.yaml
juberti: yeah, I think we'll want something like this, it's annoying that we can't pass in variables easily. For a future PR tho.

ultravox/training/train.py
juberti: Good to know.

juberti (approved):

PR #118, opened 2024-09-18
Update README: use cases for Ultravox training

farzadab:


README.md
```diff
@@ -106,23 +106,57 @@ mcli create secret gcp
...
-`torchrun --nproc_per_node=8 -m ultravox.training.train`
+1. You want to use a different LLM or audio encoder backbone.
 
-### Local Training
+    a. In this case you need to re-train the adapter. You can use `release_config.yaml`, which contains our config for our latest release, and you should be able to simply change the base LLM or encoder by specifying `--text-model <hf-model-id-for-llm>` and/or `--audio-model <hf-model-id-for-encoder>`.
 
-Here's an example command to run a training experiment using an existing config (in this case, using TinyLlama as the LLM backbone):
+2. You want to improve the knowledge of the model --> NO NEED TO TRAIN ULTRAVOX!
 
+    a. We suggest to either use RAG on the fly (no training needed), or fine-tune the LLM backbone instead. If you fine-tune the LLM, you might need to re-train Ultravox (see previous step) unless the embedding layer stays frozen.
```
zqhuang211, line 120: > unless the embedding layer stays frozen.

I am not sure about this part. I think our current training process cannot guarantee that if the embedding layer stays frozen a fine-tuned LM can be used directly with an existing adapter.

farzadab: The point of the statement wasn't that if embedding layer stays frozen you won't need to re-train, it was that if the embedding gets changed you would have to re-train.
It seems like that's not clear, so I'll make it more explicit.

farzadab: Updated sentence:
> If you fine-tune the LLM, you might still need to re-train Ultravox, for example, if the embedding layer has changed in any way.

zqhuang211: I think this statement is technically correct. However, I somewhat interpret it as "If you fine-tune the LLM and if the embedding layer stays frozen, you might not have to re-train Ultravox." I think it is slightly different from simply stating "You might need to re-train Ultravox if you fine-tune the LLM".

farzadab: You get that sense even from the new version?

farzadab: If yes, can you share how you would phrase it?

zqhuang211: I would simply write: ‘You might need to re-train Ultravox if you fine-tune the LLM.’

juberti (approved):

README.md
farzadab: Updated the text to your version.

PR #120, opened 2024-09-24
Monkey patch for HF Hub error

farzadab:
TODO: add a test


farzadab:
The upstream PR was merged: https://github.com/huggingface/huggingface_hub/pull/2567

This will likely be reflected in the next release of hub (version 0.25.2 likely), at which point all we need to do is update the dependency.

farzadab:
Fixed upstream.

PR #121, opened 2024-09-25
Dataset Tool to add Timestamps

farzadab:
This PR adds a TimestampGeneration task to `ds_tool.py`. It uses [Montreal Forced Aligner (MFA)](https://montreal-forced-aligner.readthedocs.io/en/latest/getting_started.html).

The task adds a new column (e.g. "timestamps") that has word-level alignments in the form of:
```
[ {"start": float in seconds, "end": float in seconds, "text": first word str}, ... ]
```

Usage example:
```bash
# Install Conda environment named aligner
# and download english_mfa acoustic model and dictionary
just install_mfa

# Run ds_tool (make sure upload name is not the same as input dataset_name)
just ds_tool timestamp -d fixie-ai/common_voice_17_0 -S en --upload_name fixie-ai/cv_ts  \
    -m english_mfa -T "\"{{text_proc.format_asr_text(sentence)}}\""
```

### DIFF
--- a/ultravox/tools/ds_tool/ds_tool.py
+++ b/ultravox/tools/ds_tool/ds_tool.py
@@ -2,12 +2,18 @@ import dataclasses
 import json
 import math
 import os
+import tempfile
+import subprocess
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import datasets
 import jinja2
+import librosa
 import openai
 import simple_parsing
+import soundfile as sf
+from praatio import textgrid
 import yaml
 from tenacity import retry
 from tenacity import stop_after_attempt
@@ -20,6 +26,8 @@ from ultravox.tools.ds_tool import tts
 
 tts_client: caching.CachingTtsWrapper
 chat_client: caching.CachingChatWrapper
+
+MFA_ENV_NAME = "aligner"
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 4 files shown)


ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -21,6 +29,41 @@
...
+
+    try:
+        return jinja2.Template(template, undefined=jinja2.StrictUndefined).render(
+            **sample, json_dump=json.dumps, text_proc=text_proc
+        )
+    except jinja2.TemplateError as e:
+        print(f"Error rendering template: {e}")
+        print(f"template: {template}")
+        print(f"sample keys: {list(sample.keys())}")
+        raise ValueError(
```
zqhuang211, line 63: Maybe include `exclude_fields` in the logging too to help debugging?

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +198,167 @@ def _map_sample(self, sample, exclude_fields):
...
+        writer_batch_size: int,
+        exclude_fields: List[str],
+    ) -> datasets.Dataset:
+        # 0. create a temp directory to store the audio and text files
+        _temp_dir = tempfile.TemporaryDirectory()
+        self.temp_dir = _temp_dir.name
+        os.makedirs(self.temp_dir, exist_ok=True)
+
+        # 1. copy all audio-text pairs into the temp directory
+        ds_split.map(
```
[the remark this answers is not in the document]
farzadab: Can you elaborate by what's redundant?
If you mean why we store the audio and text to file, that's just how MFA expects the inputs. I didn't find a way to invoke it from Python and let it read from the dataset, hence the awkward: store, compute, retrieve, cleanup loop.

The larger datasets use your chunking and retry machinery by default, so that's been working fine for Common Voice and has allowed for retrying failed invocations.

ultravox/tools/ds_tool/ds_tool.py
farzadab: This part of the code wasn't changed in this PR, but done.

ultravox/tools/ds_tool/ds_tool.py
```diff
@@ -175,6 +198,167 @@ def _map_sample(self, sample, exclude_fields):
...
+            self._store_sample_as_files,
+            num_proc=num_proc,
+            fn_kwargs={"exclude_fields": set(exclude_fields)},
+        )
+
+        count_wavs = len(glob.glob(os.path.join(self.temp_dir, "*.wav")))
+        assert count_wavs >= len(
+            ds_split
+        ), "Not all samples were stored as files. The id is likely not unique."
+
```
zqhuang211, line 265: what is the legitimate case when `count_waves > len(ds_split)`?

zqhuang211 (approved):

ultravox/tools/ds_tool/ds_tool.py
farzadab: I think that can only happen if we reuse the temp directory (which I was doing before). I can tighten the condition.

farzadab:
Actually if the changes here are okay I think I'll split `ds_tool` into multiple files, one per task, since this file is getting a bit too long.
I'll do it in the next PR.

ultravox/tools/ds_tool/ds_tool.py
farzadab: Done.

ultravox/tools/ds_tool/ds_tool.py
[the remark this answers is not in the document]
liPatrick: Ah okay if that's how MFA works, then ig there's no other way around it. 

I was concerned that you might run out of storage memory, so was thinking it might be good to save the audio in chunks and delete the chunk once its been processed?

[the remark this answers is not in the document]
farzadab: Oh okay yes, the temp directory's cleanup command will run an `rmtree` and removes everything inside.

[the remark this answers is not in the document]
farzadab: But that reminds me: I can put this in a `with` statement to make sure the directory is removed even if an error gets raised.

liPatrick (approved):

PR #128, opened 2024-10-05
Fix typo in README.md

ofek:


zkoch (approved):
Thanks for catching and fixing!

PR #135, opened 2024-10-16
Generic datasets with inheritance

juberti:
This is the first PR to rework our dataset handling to entirely use generic datasets. BoolQ is reworked to be entirely data-driven, and the other removed datasets will be re-added in a followup PR.

Currently, all datasets to be used are defined in the training script. This will be tweaked to have a common datasets.yaml that other YAMLs can refer to.

Some obsolete features and command line flags have been removed for simplicity.


juberti:
Tests are passing and data_tool/infer_tool all seem to be working. I'm going to update `release_config.yaml` and kick off a training run next.

ultravox/data/datasets.py
```diff
@@ -1126,24 +523,26 @@ class InterleaveDataset(SizedIterableDataset):
...
         self._datasets = datasets
         self._rng = np.random.default_rng(seed)
         self._static = static
-
         self._stop_strategy = stop_strategy
 
-        weights = [getattr(ds, "weight", 1) for ds in datasets]
+        if weights is None:
+            weights = [1.0] * len(datasets)
         total_weight = sum(weights)
```
zqhuang211, line 546: Now that all datasets have a length, so we should update the meaning of weights to represent how many copies (which can be fractional) of each dataset will be used in training. This makes it easier to configure weights in the YAML file. By default, only one copy of each dataset is used for training.

```suggestion
       weights = [w * len(d) for w, d in zip(weights, datasets)]
       total_weight = sum(weights)
```

ultravox/data/datasets.py
```diff
@@ -1206,37 +605,328 @@ class Range(SizedIterableDataset):
...
+)
+
+LS_CLEAN_CONFIG = DatasetConfig(
+    base="librispeech",
+    subset="clean",
+    splits=[
+        DatasetSplitConfig(name="train.100", num_samples=28_539),
+        DatasetSplitConfig(name="train.360", num_samples=104_014),
+    ],
+    user_template_args={"foo": "bar"},
```
zqhuang211, line 726: Do we need to keep this?

ultravox/data/datasets.py
```diff
@@ -1206,37 +605,328 @@ class Range(SizedIterableDataset):
...
+    "peoplespeech": PS_CLEAN_CONFIG,
+    "peoplespeech-transcription": PS_CLEAN_TRANS_CONFIG,
+    "peoplespeech-continuation": PS_CLEAN_CONT_CONFIG,
+    "voxpopuli": VP_EN_CONFIG,
+}
+
+
+def register_datasets(datasets: Dict[str, DatasetConfig]):
+    for name, config in datasets.items():
+        DATASET_MAP[name] = config
```
zqhuang211, line 908: should we do sanity check to avoid accidentally overriding datasets?

ultravox/data/datasets.py
```diff
@@ -8,70 +8,117 @@
...
 
-
 # Silence the spurious warnings coming from the MosaicML streaming library.
 logging.getLogger("streaming.base.dataset").setLevel(logging.ERROR)
 
 
+class DatasetSplit(str, enum.Enum):
+    TRAIN = "train"
+    VALIDATION = "validation"
+
```
zqhuang211, line 40: I assume that we could also include a `evaluation` split, for example, for Speech Translation evaluation?

ultravox/data/datasets.py
```diff
@@ -8,70 +8,117 @@
...
+
+@dataclasses.dataclass
+class VoiceDatasetArgs:
+    """Global arguments for voice datasets."""
+
+    batch_size: int = 4
+    """Batch size for train, eval, or validation."""
+    include_audio: bool = True
```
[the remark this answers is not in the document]
zqhuang211: Right now `VoiceDatasetArgs` is not directly configured in the YAML file but instead in the `train.py` file with some selected fields configurable in the  YAML file. I assume we could in the future make separate `VoiceDatasetArgs` for training, evaluation, and valuation datasets, right?

zqhuang211:
> This is the first PR to rework our dataset handling to entirely use generic datasets. BoolQ is reworked to be entirely data-driven, and the other removed datasets will be re-added in a followup PR.
> 
> Currently, all datasets to be used are defined in the training script. This will be tweaked to have a common datasets.yaml that other YAMLs can refer to.
> 
> Some obsolete features and command line flags have been removed for simplicity.

Overall this looks very good. Thank you!

ultravox/data/datasets.py
```diff
@@ -289,134 +300,116 @@ class VoiceDataset(SizedIterableDataset):
...
-    def _load_audio_dataset(
+    def _load_hf_dataset(
         self,
         path: str,
         name: Optional[str] = None,
         *,
         split: Optional[str] = None,
-        shuffle: Optional[bool] = None,
         streaming: bool = True,
+        audio_field: Optional[str] = None,
```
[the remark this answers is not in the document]
juberti: Yeah, the real challenge here is figuring out which audio to load in `_get_audio` since we don't have a mechanism to select that on a sample-by-sample basis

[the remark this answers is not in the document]
juberti: Accordingly, will plan to leave as-is for now.

ultravox/data/datasets.py
```diff
@@ -289,134 +300,116 @@ class VoiceDataset(SizedIterableDataset):
...
+
+            yield sample
             actual_length += 1
-            # If len(dataset) == 0 most likely the dataset is a validation dataset,
-            # or the training is using max_steps instead of num_epochs.
-            if actual_length > len(self) and len(self) > 1:
+            if actual_length == len(self) + 1:
                 warnings.warn(
-                    f"The estimated length {self._estimated_length} has been exceeded for type {type(self._dataset)}. Make sure to update."
+                    f"The presumed length {self._length} has been exceeded for split {self._dataset.split}. Make sure to update."
```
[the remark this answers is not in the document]
juberti: Good question, will look into this

ultravox/data/datasets.py
juberti: oops, debugging code, will remove

ultravox/data/datasets.py
juberti: good idea

ultravox/data/datasets.py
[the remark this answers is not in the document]
juberti: Good suggestion. I don't really want to duplicate each dataset to get a text-only version, but I think having per-dataset `VoiceDatasetArgs` might be possible.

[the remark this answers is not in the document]
farzadab: Hmm, interesting.
One main reason for putting this in `main.py` was text-only validations (i.e. to be able to override it from `train.py`). I don't think we use them as often though.

ultravox/data/datasets.py
```diff
@@ -1206,37 +605,328 @@ class Range(SizedIterableDataset):
...
+BOOLQ_CONFIG = DatasetConfig(
+    path="fixie-ai/boolq-audio",
+    splits=[
+        DatasetSplitConfig(name="train", num_samples=10000),
+        DatasetSplitConfig(name="validation", num_samples=1000),
+    ],
+    user_template="{{passage}}\n\n{{'<|audio|>' if include_audio else question}}",
+    assistant_template="{{'True' if answer else 'False'}}",
+    transcript_template="{{question}}",
+)
```
farzadab, line 640: I'm wondering if we should put all of these dataset definitions in a `base_datasets.yaml` file. That way this file becomes way smaller too and only contains logic.

We just have to register them (e.g. at the end of this file) and we might need some variables, which IIRC yaml already has.

ultravox/data/datasets.py
```diff
@@ -8,70 +8,117 @@
...
-    "First listen to the clip. Then, transcribe exactly what is said. <|audio|>",
-    # from https://arxiv.org/pdf/2402.08846
-    "Transcribe speech to text: <|audio|>",
-    # from GPT-4
-    "Capture every word from the audio verbatim\n<|audio|>",
-    "Convert speech to text from audio\n<|audio|>",
-    "Listen and transcribe the complete text from audio\n<|audio|>",
-    "Record in writing what is spoken in audio\n<|audio|>",
-    "Transcribe the spoken words from audio with exact wording and punctuation\n<|audio|>",
-]
```
farzadab, line 50: I know we haven't been using these, but just making sure: @zqhuang211 did you do an experiment to show that randomizing the prompt is not useful at all?

farzadab (commented):
This looks great, thanks!
Added some comments. I'll do another pass of review, but it LGTM.

ultravox/training/train.py
```diff
@@ -197,69 +199,51 @@ def train(args: config_base.TrainConfig):
...
+        val_dataset = prepare_dataset(
+            train_args=args,
+            data_sets_and_weights=args.val_sets,
+            train_on_inputs=args.train_on_inputs,
+            stop_strategy=args.stop_strategy,
+            processor=processor,
+            num_samples=args.val_num_samples,
+            data_args=val_ds_args,
+            include_alt_fields=model.loss_config.requires_alt_fields,
+        )
```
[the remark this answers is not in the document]
farzadab: I think you can remove the `text_*` version of the validations since they never change.

ultravox/training/configs/llama_whisper.yaml
```diff
@@ -0,0 +1,103 @@
...
+exp_name: "llama3.1-8b-whisper"
+text_model: "meta-llama/Meta-Llama-3.1-8B-Instruct"
+audio_model: "openai/whisper-medium"
+
+loss_config:
+  loss_function: "KL_Divergence"
+
+max_steps: 20 # x8x24 = 2,764,800
+
+# This would go in a datasets.yaml file and we could either use pyyaml-include to include it
```
farzadab, line 12: Oh nice, I was looking for a way to include YAML files, assuming it works with `simple_parsing`.

ultravox/training/configs/llama_whisper.yaml
```diff
@@ -0,0 +1,103 @@
...
+data_sets:
+  librispeech:
+    path: "fixie-ai/librispeech_asr"
+    user_template: "<|audio|>"
+    assistant_template: ""
+    transcript_template: "{{ text }}"
+
+  # Note the inheritance here
+  librispeech-clean:
+    base: "librispeech" # this could also be done via "<<": *librispeech, although that approach is less flexible
```
farzadab, line 24: Agreed. YAML variables are probably good for small-scale changes, but for this I like your approach more.

ultravox/data/datasets.py
```diff
@@ -1206,37 +605,328 @@ class Range(SizedIterableDataset):
...
+    while temp:
+        config = DATASET_MAP[temp]
+        configs.insert(0, config)
+        temp = config.base
+    # Set the root config, and then apply any non-None overrides from the subclasses.
+    merged_config = dataclasses.replace(configs[0])
+    for config in configs[1:]:
+        for field in dataclasses.fields(config):
+            value = getattr(config, field.name)
+            if field.name != "base" and value is not None:
```
farzadab, line 925: I guess this method won't work recursively (e.g. for `splits` and `user_template_args`) which is okay. That's more than fine for `splits`, but a bit tricky with `user_template_args`.

This is all fine, but it would be good to put a comment in a few places though (e.g. one here and one in `DatasetConfig`).

farzadab (approved):
A few more comments.

ultravox/data/datasets.py
```diff
@@ -470,654 +439,82 @@ def _make_sample(
...
-        dataset = self._load_audio_dataset(
-            "fixie-ai/peoples_speech", "clean", split=args.split.value
-        )
-        self._init_dataset(dataset)
-
-    def _get_sample(self, row) -> Optional[VoiceSample]:
-        return self._get_transcribe_sample(row, tcol="text")
-
-
-class SodaDataset(VoiceDataset):
```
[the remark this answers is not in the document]
juberti: OK, I added an initial stab at this but the work required to get everything working is outside the scope of this PR.

ultravox/data/datasets.py
```diff
@@ -8,70 +8,117 @@
...
-    "Respond to this question: \n<|audio|>",
-    "Continue the conversation after <|audio|>",
-    "First listen to the clip: <|audio|>\n How would you respond?",
-    "<|audio|> - respond",
-    "<|audio|>\n Respond to the question",
-]
-
 # TODO(juberti): set these in the environment so they don't need to be hard-coded here.
 os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = "service_account.json"
 os.environ["GOOGLE_CLOUD_PROJECT"] = "fixie-training"
```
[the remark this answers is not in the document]
juberti: We don't right now. If the change here to do retries addresses the connection errors then I think we can remove this in a future PR.

ultravox/data/datasets.py
juberti: yep, we can definitely do that. Would that typically be the "test" split?

ultravox/data/datasets.py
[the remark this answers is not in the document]
juberti: I tweaked the structure of `train_sets` and `val_sets` to allow arbitrary parameters to be included, e.g., `include_audio`, which can be fed to the model as dataset-specific `VoiceDatasetArgs`. Some more work will be needed to fully wire this up, which I can take on in a future PR.

ultravox/data/datasets.py
juberti: I was fiddling with this and I think it can be factored out, but the imports started to get out of hand, will revisit in a future PR.

ultravox/training/train.py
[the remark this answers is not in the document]
juberti: I was trying to simplify things a bit as I saw comments indicating that some stuff here wasn't working. I can definitely return a dict easily instead though.

ultravox/data/datasets.py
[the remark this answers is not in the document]
juberti: yeah, it's None in that case. I reworked this.

ultravox/training/train.py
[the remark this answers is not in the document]
farzadab: Yeah I think that would be best, thanks. I'd not add the text_* and matchtrain back though.

ultravox/data/datasets.py
juberti: I think this is a more comprehensive change to `InterleavedDataset`, as the `len()` method will need to be adjusted, and the stop strategy changed to run until the overall `len()` is reached. I don't think it's a ton of work but should be done in a separate PR.

ultravox/data/datasets.py
juberti: We could bring this functionality back by letting `user_template` be a List, if needed.

farzadab: Agreed. I don't want to block on this.

ultravox/training/train.py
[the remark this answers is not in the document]
juberti: ok, done

juberti:
all comments addressed or responded to, PTAL

juberti:
Training with release_config.yaml now, https://wandb.ai/fixie/ultravox/runs/e0q33e2f

zqhuang211 (approved):

farzadab (approved):
LGTM. I especially love that the number of lines deleted is more than the lines added (not joking) 😁

PR #140, opened 2024-10-18
Switch InterleaveDataset to use weights (e.g., 2.0, 0.5, etc)

juberti:
This makes it easy to control how many epochs a given dataset is used for, and makes interleaving always deterministic (we always vend from the least-used dataset).

As part of this I removed the stop-strategy member, as it doesn't really make sense in this new context. LMK if I'm misunderstanding things.

Related to #105


liPatrick (approved):
lgtm, and I think if we need to extend this to use the stop strategy in the future your approach lets us do that pretty easily.

ultravox/data/datasets.py
```diff
@@ -550,77 +549,48 @@ def __len__(self):
...
+        ds_iters = [iter(ds) for ds in self._datasets]
+        ds_pos = [0] * len(ds_iters)
+        # Find the iterator that is least far along and vend from it.
+        for i in range(self._total_samples):
+            min_fraction = 1.0
+            for j in range(len(ds_iters)):
+                iter_fraction = ds_pos[j] / self._weighted_samples[j]
+                if iter_fraction < min_fraction:
+                    min_fraction = iter_fraction
+                    iter_index = j
```
farzadab, line 584: nit: this can be computed faster, but probably not worth it.

ultravox/data/datasets.py
```diff
@@ -550,77 +549,48 @@ def __len__(self):
...
-        while True:
-            if self._static:
-                iter_index = next(static_iter)
-            else:
-                iter_index = self._rng.choice(len(iters), p=self._normalized_probs)
-
+        ds_iters = [iter(ds) for ds in self._datasets]
+        ds_pos = [0] * len(ds_iters)
+        # Find the iterator that is least far along and vend from it.
+        for i in range(self._total_samples):
```
farzadab, line 578: This for loop can cause issues.

I believe this method is gonna work fine if the weights are whole numbers, but with fractions and specially `weight < 1` I'm not so sure:

`ds_iters` will get re-started when the for loop ends and:
1. Unless we reshuffle the dataset, you're always going to get the same subset. That means if you set `weight=0.1`, you're never going to see 90% of the data.
2. Even with reshuffling, since the iterator gets reset before you do a full epoch, you might see some samples multiple times while some samples haven't been seen yet.

farzadab: See this for item 1:
```python
>> ds = datasets.load_dataset('librispeech_asr', split="validation.clean", streaming=True)
>> ds = ds.shuffle()
>>> next(iter(ds))
{'file': '6345-64257-0015.flac', 'audio': {'path': '6345-64257-0015.flac', 'array': array([-0.00195312, -0.00149536, -0.00149536, ...,  0.00015259,
        0.00054932,  0.00143433]), 'sampling_rate': 16000}, 'text': 'I KNOW IT AND THERE IS NO WAKING', 'speaker_id': 6345, 'chapter_id': 64257, 'id': '6345-64257-0015'}
>>> next(iter(ds))
{'file': '6345-64257-0015.flac', 'audio': {'path': '6345-64257-0015.flac', 'array': array([-0.00195312, -0.00149536, -0.00149536, ...,  0.00015259,
        0.00054932,  0.00143433]), 'sampling_rate': 16000}, 'text': 'I KNOW IT AND THERE IS NO WAKING', 'speaker_id': 6345, 'chapter_id': 64257, 'id': '6345-64257-0015'}
 ```

Comment (juberti) [reply ultravox/data/datasets.py:578]: Agreed. In a sense the weight is really just the number of "epochs" to run for the dataset, so from that view the behavior matches expectations (for both 0.1 and 1.1). Is there a situation where you'd see the same 0.1 of samples multiple times?

If needed, one easy thing we could do would be to add a random offset to each dataset. For 1.0 you would get the same samples just in a unique order, but for 0.1 you'd get a random 10% slice.

File: ultravox/data/datasets.py (hunk 1 above)
Comment (zqhuang211) [reply ultravox/data/datasets.py:584]: I agree that we should remove the logic involving `stop_strategy`, but we could use still use the random sampling strategy based on the following line:

```
self._weighted_samples = [int(w * len(d)) for w, d in zip(weights, datasets)]
self._normalized_probs = [ w / self._total_samples for w in self._weighted_samples ] 
iter_index = self._rng.choice(len(iters), p=self._normalized_probs)
```
This would ensure that on average the samples following the datasets would follow the distribution according to `self._normalized_probs`.

Comment (juberti) [reply ultravox/data/datasets.py:584]: But we'll get that with the new code even without the random sampling. It will follow the probabilities in deterministic fashioin.

File: ultravox/data/datasets.py (hunk 2 above)
Comment (farzadab) [reply ultravox/data/datasets.py:578]: Not sure I understood what you said:

> Is there a situation where you'd see the same 0.1 of samples multiple times?

Is that a question of do we ever want this? The answer is I think no. I don't see a real advantage in having that. Maybe just for testing :?

> one easy thing we could do would be to add a random offset to each dataset

That's kinda similar to what I suggested with shuffling and so the issue would be the same: with `weight=0.1`, if you do 10 epochs, you're gonna see some samples multiple times and some samples 0 times.

File: ultravox/data/datasets.py (hunk 1 above)
Comment (zqhuang211) [reply ultravox/data/datasets.py:584]: It is not clear to me what the benefit is to follow probabilities in deterministic fashion. If we want create random batches, we should allow random sampling of subsets, regardless of whether it would materially change the quality of the trained model. A deterministic order can be achieved by choosing the seed of the random generator.

I am also okay with using a deterministic order. Just thought that there is a cleaner alternative.

Comment (juberti) [reply ultravox/data/datasets.py:584]: But the randomness here is just the iterator selection, i.e., what dataset is sampled from next. This adds a small amount of noise but still overall selects the same samples. What's the real value in that?

File: ultravox/data/datasets.py [hunk 3]
```diff
@@ -550,77 +549,48 @@ def __len__(self):
...
-            static_iter = itertools.cycle(range(len(self._datasets)))
-
-        while True:
-            if self._static:
-                iter_index = next(static_iter)
-            else:
-                iter_index = self._rng.choice(len(iters), p=self._normalized_probs)
-
+        ds_iters = [iter(ds) for ds in self._datasets]
+        ds_pos = [0] * len(ds_iters)
```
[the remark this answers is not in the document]
farzadab: That probably works in practice, but it breaks the interface (since different people might call `iter` in different places) and if there's a bug it's gonna be hard to catch.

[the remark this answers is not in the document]
juberti: Let's take a step back first: what behavior do we think is correct here?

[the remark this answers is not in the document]
zqhuang211: Farzad has a fair point. Here are the requirements I have in mind:

1. Already met:
   a) Shuffling of samples within individual datasets.
   b) The number (or distribution) of selected samples from each dataset should align with the specified weight for an epoch (or at any point during training).
2. Not yet met:
   a) Effective use of all samples from a dataset
       Although weight is specified as the percentage of samples used for training in an epoch, it actually reflects the relative importance of the datasets. Consider datasets A and B, each with 1M samples, where A is noisier than B. We may assign weights of 0.1 to A and 1.0 to B, and train for 3 epochs. In this case, we definitely want a different subset of samples from A to be used in the second epoch, and similarly for the third.
   b) New shuffling of samples when a dataset is iterated in subsequent epochs of the dataset (not the specified training epochs)


Regarding the difference between deterministic and random sample, it seems to me that the main advantage of the deterministic approach is to make sure that the exact number of samples from each dataset is used in an epoch, and the distribution of selected samples matches the specified distribution as close as possible at any point of training. These are nice attributes to have.

ultravox/data/datasets.py
farzadab: I agree that deterministic and random methods are the same. The difference maker is the change from "for loop" to "while True till exhausted".

I do think that the "while True" method had other issues (e.g. the distribution of data will change when one dataset is exhausted before others). I assume that ZQ had accepted that trade-off with the previous code.

zqhuang211: I was mainly thinking from the angle of why we need to change the 'random sampling' method to the 'deterministic' method. I think I may have answered the question in the other post.

PR #141, opened 2024-10-19
Break up datasets.py

juberti:
This splits out types.py and registry.py to move the list of pre-defined datasets to its own file and avoid circular refs.

An __all__ import is used to minimize changes to surrounding code.

### DIFF
--- a/ultravox/data/registry.py
+++ b/ultravox/data/registry.py
@@ -1,0 +1,397 @@
+import dataclasses
+from typing import Dict, List, Optional
+
+from ultravox.data import datasets
+from ultravox.data import types
+
+CONTINUATION_USER_TEMPLATE = f"Continue the following text using less than 50 words:\n\n{types.AUDIO_PLACEHOLDER}"
+CONTINUATION_ASSISTANT_TEMPLATE = "{{continuation}}"
+TRANSCRIPTION_USER_TEMPLATE = f"Transcribe\n{types.AUDIO_PLACEHOLDER}"
+
+BOOLQ_CONFIG = types.DatasetConfig(
+    name="boolq",
+    path="fixie-ai/boolq-audio",
+    splits=[
+        types.DatasetSplitConfig(name="train", num_samples=10000),
+        types.DatasetSplitConfig(name="validation", num_samples=1000),
+    ],
+    user_template="{{passage}}\n\n{AUDIO_PLACEHOLDER}",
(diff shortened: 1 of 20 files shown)


ultravox/data/types.py
```diff
@@ -0,0 +1,206 @@
...
+    return base64.b64encode(audio_to_wav(audio, sample_rate)).decode("utf-8")
+
+
+def audio_to_data_uri(audio: np.ndarray, sample_rate: int = SAMPLE_RATE) -> str:
+    """Convert audio to a data URI."""
+    return f"data:audio/wav;base64,{audio_to_wav_base64(audio, sample_rate)}"
+
+
+def messages_from_prompt(prompt: str) -> List[Dict[str, str]]:
+    return [{"role": "user", "content": prompt}]
```
zqhuang211, line 141: Should we define these methods in a separate file, like `helpers.py`, as they are not data `types`.

ultravox/data/registry.py
```diff
@@ -0,0 +1,397 @@
...
+    name="voxpopuli-en",
+    path="facebook/voxpopuli",
+    subset="en",
+    splits=[
+        types.DatasetSplitConfig(name="train", num_samples=1_000_000),
+        types.DatasetSplitConfig(name="validation", num_samples=10_000),
+    ],
+    assistant_template="{{raw_text}}",
+    transcript_template="{{raw_text}}",
+)
```
zqhuang211, line 164: should `voxpopuli-en` be defined in a similar way as `commonvoice-en`, starting from a base config named `voxpopuli`?

zqhuang211 (approved):

ultravox/data/types.py
juberti: yeah, fair point, I can move this to a new file like `sample.py`.

ultravox/data/registry.py
```diff
@@ -0,0 +1,397 @@
...
+    transcript_template="{{question}}",
+)
+
+CV_BASE_CONFIG = types.DatasetConfig(
+    name="commonvoice",
+    path="fixie-ai/common_voice_17_0",
+    assistant_template="{{sentence}}",
+    transcript_template="{{sentence}}",
+)
+
```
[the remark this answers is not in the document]
juberti: The registry could load itself from a datasets.yaml file. I'll probably do that in a separate PR though.

PR #146, opened 2024-11-07
Add whisper masking

zqhuang211:
- Added masking in whisper encoder to ensure consistency in training and inference.
- Simplified `release_config.yaml` to serve as an example configuration.


liPatrick (commented):
Doesn't whisper expect us to pad to 30sec? Do we have any concern about masking the padding here? Other than that the code for the pad masking looks good

liPatrick (approved):

zqhuang211:
> Doesn't whisper expect us to pad to 30sec? Do we have any concern about masking the padding here? Other than that the code for the pad masking looks good

Yes, Whisper is trained with padding to 30 seconds, so the way we’ve used it may cause some mismatch. However, it does not seem to degrade end-to-end performance in our (limited) comparative studies.

PR #148, opened 2024-11-09
Audio streaming training with masking

saeeddhqan:
By utilizing a mask, we can simulate audio streaming.

### DIFF
--- a/ultravox/model/ultravox_model.py
+++ b/ultravox/model/ultravox_model.py
@@ -548,7 +548,7 @@ class ModifiedWhisperEncoder(
         if audio_latency_block_size is None:
             self.audio_streaming_mask = None
             return
-        audio_streaming_mask = None
+
         # maximum sequence length
         max_seqlen = (
             self.config.max_source_positions
--- a/ultravox/model/ultravox_model.py
+++ b/ultravox/model/ultravox_model.py
@@ -55,6 +55,24 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
             self.audio_tower._no_split_modules or []
         )
 
+        audio_latency_bsize = config.audio_latency_bsize
+        audio_streaming_mask = None
+        if audio_latency_bsize != -1:
+            audio_latency_nblocks = (self.audio_tower.config.max_source_positions
+                * self.audio_tower.conv1.stride[0]
+                * self.audio_tower.conv2.stride[0]
+            )
+            audio_latency_nblocks //=  2 * audio_latency_bsize
+            audio_streaming_mask = torch.tril(
+                torch.ones(
+                    audio_latency_nblocks, audio_latency_nblocks),
+                    diagonal=0,
+                ).repeat_interleave(audio_latency_bsize, dim=0).repeat_interleave(audio_latency_bsize, dim=1)
+            audio_streaming_mask[audio_streaming_mask == 0] = float('-inf')
+            audio_streaming_mask[audio_streaming_mask == 1] = 0
+            audio_streaming_mask = audio_streaming_mask[None, None, :, :]
+        self.register_buffer('audio_streaming_mask', audio_streaming_mask, persistent=False)
+
         self.loss_config = LossConfig()
         self.post_init()
 
@@ -190,7 +208,9 @@ class UltravoxModel(transformers.LlamaPreTrainedModel):
 
             # B x A/3200 x D
             audio_tower_output = self.audio_tower.forward(
-                audio_values.to(self.audio_tower.dtype), audio_len=audio_len
+                audio_values.to(self.audio_tower.dtype),
+                audio_len=audio_len,
+                attention_mask_bw=self.audio_streaming_mask,
             ).last_hidden_state
             audio_tower_output = audio_tower_output.to(inputs_embeds.dtype)
 
@@ -537,6 +557,7 @@ class ModifiedWhisperEncoder(
         output_attentions=None,
         output_hidden_states=None,
         return_dict=None,
+        attention_mask_bw=None,
     ):
         expected_seq_length = (
             self.config.max_source_positions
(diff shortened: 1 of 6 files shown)


ultravox/model/ultravox_model.py
```diff
@@ -55,6 +55,24 @@ def __init__(self, config: UltravoxConfig):
...
         )
 
+        audio_latency_bsize = config.audio_latency_bsize
+        audio_streaming_mask = None
+        if audio_latency_bsize != -1:
+            audio_latency_nblocks = (self.audio_tower.config.max_source_positions
+                * self.audio_tower.conv1.stride[0]
+                * self.audio_tower.conv2.stride[0]
+            )
+            audio_latency_nblocks //=  2 * audio_latency_bsize
```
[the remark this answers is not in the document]
farzadab: Actually, if `max_source_positions`  is not divisible by `audio_latency_bsize`, we might be in trouble, so I'd rather put an `assert` here.

farzadab (commented):
Thanks Saeed. I left some comments.

It would be ideal to add some tests, for example to make sure the structure of the mask matches what we expect. If you can think of another test feel free to add one.
I know the repo doesn't have many tests, but we'd like to change that.

saeeddhqan:
@farzadab Hi,
I created a test file for ultravox_model, moved the function, and changed var names. 
I look forward to further improvements.

ultravox/model/ultravox_model.py
```diff
@@ -286,13 +287,16 @@ def _create_audio_tower(
                 audio_tower = ModifiedWhisperEncoder.from_pretrained(
                     config.audio_model_id, torch_dtype=config.torch_dtype
                 )
+                audio_tower.init_latency_mask(
+                    config.audio_latency_block_size, dtype=config.torch_dtype)
```
farzadab, line 291: Thanks. It would be good to have an assert in the other case as well if block_size is non-zero rather than silently refusing to use it.

ultravox/model/ultravox_model.py
```diff
@@ -529,6 +533,35 @@ class ModifiedWhisperEncoder(
     base_model_prefix = "model.encoder"
     _no_split_modules = ["WhisperEncoderLayer"]
 
+    def init_latency_mask(self, audio_latency_block_size: int, dtype: torch.float):
+        if audio_latency_block_size is None:
+            self.audio_streaming_mask = None
+            return
+        audio_streaming_mask = None
+        # total sequence length
```
farzadab, line 541:
```suggestion
        # maximum sequence length
```

ultravox/model/ultravox_model.py
```diff
@@ -529,6 +533,35 @@ class ModifiedWhisperEncoder(
...
+
+        audio_latency_nblocks //=  2 * audio_latency_block_size
+        audio_streaming_mask = torch.tril(
+            torch.ones(
+                audio_latency_nblocks, audio_latency_nblocks),
+                diagonal=0,
+            ).repeat_interleave(audio_latency_block_size, dim=0).repeat_interleave(audio_latency_block_size, dim=1)
+        audio_streaming_mask = (1.0 - audio_streaming_mask) * torch.finfo(dtype).min
+        # audio_streaming_mask[audio_streaming_mask == 0] = float('-inf')
+        # audio_streaming_mask[audio_streaming_mask == 1] = 0
```
farzadab, line 561: nit: please remove commented code

ultravox/training/configs/streaming_tinyllama.yaml
```diff
@@ -0,0 +1,26 @@
+# SLM with ultravox & llama3.1, trained wtih knowledge distillation.
```
farzadab, line 1: Do you have any experimental results with this in Ultravox repo?
Also the experiment name mentions "1s" whereas the `block_size` is `null` which seems incorrect.

ultravox/model/ultravox_model_test.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+import torch
+from transformers import WhisperConfig
+from transformers.models.whisper import modeling_whisper as whisper
+from ultravox.model import ultravox_model
+
+class TestModifiedWhisperEncoder:
```
farzadab, line 7: nit: We don't use classes for our tests. It's probably best to use stand-alone functions for consistency since I don't see the `self` being useful anywhere.

ultravox/model/ultravox_model.py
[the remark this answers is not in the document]
farzadab: I don't think this was answered. The `*2` is still there and I don't see an explanation here or in the code about it. It would be ideal to avoid magic numbers as much as possible.

farzadab (commented):
Thanks. I left a few nit comments.
Please make sure to run `just format` to run formatting and use `just test` to make sure the tests pass.

ultravox/training/configs/streaming_tinyllama.yaml
saeeddhqan: Yes, I did all the experiments with the forked repo; and I normally run a training with a few steps after conducted changes to make sure things work smoothly.

saeeddhqan:
@farzadab, I utilized 'just test', and 'just format', and corrected mentioned issues.

farzadab (approved):
Assuming the tests pass it LGTM. Thanks Saeed!

ultravox/training/config_base.py
```diff
@@ -106,6 +106,9 @@ def get_val_sets(self) -> List[DatasetOptions]:
     # loss function to use
     loss_config: Optional[ultravox_config.LossConfig] = None
 
+    # To simulate audio streaming with masking. None for non-causal, 100 for 1s, 200 for 2s, and so on.
+    audio_latency_block_size: int = None
```
farzadab, line 110:
```suggestion
    audio_latency_block_size: Optional[int] = None
```

ultravox/model/ultravox_model.py
```diff
@@ -529,6 +544,39 @@ class ModifiedWhisperEncoder(
     base_model_prefix = "model.encoder"
     _no_split_modules = ["WhisperEncoderLayer"]
 
+    def init_latency_mask(self, audio_latency_block_size: int, dtype: torch.float):
```
farzadab, line 547: Is this what you meant?
```suggestion
    def init_latency_mask(self, audio_latency_block_size: int, dtype: torch.dtype):
```

farzadab (commented):
Seems like `just test` failed since `mypy` is complaining.

I proposed the solution for two of the issues. The last one I couldn't figure out at a glance why the value could be None, but it should be easy to figure out with the code editor's help.

saeeddhqan:
@farzadab solved.

PR #160, opened 2024-11-27
Fix processor being overwritten by parent class

farzadab:
In the most recent versions of `transformers`, the `Pipeline` parent class overrides `self.processor` and sets it to `None`. This PR side-steps that by changing the order in which `self.processor` is assigned.


farzadab:
Already applied. Closing.

PR #280, opened 2025-02-10
Training Stability: Patch HF Hub and Datasets methods and update datasets.py

farzadab:
We've experienced some issues when training with the basic HF ecosystem. This PR adds some monkey-patches on HF methods (in HF Hub and Datasets) to
1. alleviates issues like time outs or transient connection errors when downloading data
2. make datasets more robust to errors like audio field being empty which was not recoverable without patching

As well as some minor cosmetic changes.


zqhuang211 (approved):
LGTM

PR #310, opened 2025-07-29
Allow response generation with no user message

matthewclso:
I may be wrong, but it is currently not possible to generate an assistant message without first inputting a user message.  This would allow for situations where you might want to generate an assistant message with just a system message prompt.  This PR allows for `sample` to be `None`, provided there are previous messages in the history (such as system messages).  I am not sure if this is the best implementation - so please let me know if it needs any changes!


zqhuang211 (approved):
It looks good. Thank you!

PR #312, opened 2025-08-08
Add voicebench evaluation suite

zqhuang211:
Integrate voicebench evaluation suite, see https://github.com/MatthewCYM/VoiceBench


ultravox/data/datasets.py
```diff
@@ -344,6 +353,17 @@ def __str__(self):
 
     def _get_sample(self, row) -> Optional[data_sample.VoiceSample]:
 
+        # Setting up extra_kwargs for datasets like Voicebench
+        extra_kwargs = None
+        if (
```
ravi-fixie, line 358: Did you change the eval_config to None in post_init function in types.py? It used to be  an empty dict which could make self._config.eval_config.extra_kwargs_map fail.

ultravox/data/types.py
```diff
@@ -201,7 +204,7 @@ def __post_init__(self):
             "audio_field": "audio",
             "use_mds": False,
             "mds_batch_size": 32,
-            "eval_config": {},
+            "eval_config": None,
```
ravi-fixie, line 207: There it is. ignore my previous comment.

ravi-fixie (approved):

ultravox/data/datasets.py
zqhuang211: Yes, it should be None instead of `{}`, which is not a valid `Optional[EvalConfig]` type and would cause failures in `just test`

ravi-fixie (approved):
Apart from the eval_config change, everything appears to be same.

PR #319, opened 2025-10-16
feat(models): add argument to gradio to allow quantisation for the model

Alexandre-nk-Perdereau:
This PR adds optional quantization support for Ultravox models using bitsandbytes, controlled by a new `load_in_nbit` parameter (accepts 4 or 8).

Tested on ultravox-v0_6-llama-3_1-8b:
- **Without quantization**: 17.232 GB VRAM
- **8-bit quantization**: 10.510 GB VRAM
- **4-bit quantization**: 8.314 GB VRAM

Only the LLM is quantized using bitsandbytes. The audio tower and multi-modal projector remain in float16 to preserve audio processing quality and avoid potential degradation in audio understanding.

The quantization is implemented through a custom loader (quantized_loader.py) because Ultravox's composite architecture (language_model + audio_tower + projector) doesn't work well with standard "from_pretrained" quantization. Each component is loaded separately with appropriate settings.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -635,6 +635,29 @@ test-tox = ["equinox", "jax[cpu]", "jaxtyping", "mypy (>=0.800)", "numba", "nump
 test-tox-coverage = ["coverage (>=5.5)"]
 
 [[package]]
+name = "bitsandbytes"
+version = "0.48.1"
+description = "k-bit optimizers and matrix multiplication routines."
+optional = false
+python-versions = ">=3.9"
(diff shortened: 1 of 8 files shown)


Alexandre-nk-Perdereau:
I have identified a problem where this quantization makes the results of fixie-ai/ultravox-v0_6-qwen-3-32b completely erratic. 

I don't have enough VRAM to run this model without quantization, can someones confirms it works well without the quantization ?

